PR #71, opened 2016-03-30
Fix cacerts generation from pem.

ijmad:
There is no need for the private keys.
cacerts is a truststore, not a keystore.

### DIFF
--- a/docker-startup.sh
+++ b/docker-startup.sh
@@ -4,11 +4,4 @@ if [ "${ENABLE_NEWRELIC}" == "yes" ]; then
   NEWRELIC_JVM_FLAG="-javaagent:/app/newrelic/newrelic.jar"
 fi
 
-# (re)create cacerts from supplied certificates
-rm /app/ssl/cacerts 
-for crt in `ls -1 $CA_FILEPATH/*.crt`; do
-  echo "Found $crt"
-  keytool -import -file $crt -alias $crt -keystore /app/ssl/cacerts -storepass password -noprompt
-done
-
 java ${NEWRELIC_JVM_FLAG} -jar *-allinone.jar server *.yaml
--- a/docker-startup.sh
+++ b/docker-startup.sh
@@ -1,35 +1,14 @@
 #!/usr/bin/env bash
 
-#create keystore and import certs if necessary properties present
-function configure_keystore {
-    if [ -z "${CERTS_DIR}" -a \
-         -z "${KEYSTORE_DIR}" -a \
-         -z "${KEYSTORE_PASSWORD}" -a \
-         -z "${KEYSTORE_FILE}" -a \
-         -z "${CERT_FILE}" -a \
-         -z "${KEY_FILE}" ];
-         then
-           echo "not all variables needed for keystore creation are available";
-         else
-           if [ -d "${CERTS_DIR}" -a -d "${KEYSTORE_DIR}" ];
-           then
-                #replace correct filenames
-                openssl pkcs12 -export -out ${KEYSTORE_DIR}/${KEYSTORE_FILE} -inkey ${CERTS_DIR}/${KEY_FILE} -in ${CERTS_DIR}/${CERT_FILE} -passout pass:${KEYSTORE_PASSWORD};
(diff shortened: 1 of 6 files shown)


ijmad:
Hold off on this for a sec, I am pondering approaches that involve loading the pems programmatically and avoiding the need for bash scripting as a result.

src/main/java/uk/gov/pay/api/utils/TrustStoreLoader.java
```diff
@@ -0,0 +1,64 @@
...
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Paths;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+
+public class TrustStoreLoader {
```
kakumara: Isn't this a bit of an overkill for this task? 
  - We should try and avoid re-loading certs if they are already loaded. This might load it everytime the container starts up.   
 - My thoughts about this was Perhaps bash might be a better option in this case? And remove the certs after a successful load.

docker-startup.sh
```diff
@@ -1,35 +1,14 @@
...
   NEWRELIC_JVM_FLAG="-javaagent:/app/newrelic/newrelic.jar"
 fi
 
-## TODO enable this after figuring out the failure during `openssl` above
-# configure_keystore
+# (re)create cacerts from supplied certificates
+rm /app/ssl/cacerts 
+for crt in `ls -1 $CA_FILEPATH/*.crt`; do
+  echo "Found $crt"
+  keytool -import -file $crt -alias $crt -keystore /app/ssl/cacerts -storepass password -noprompt
```
kakumara: The reason I didn't use keytool for this was that it doesn't support `pkcs 12` format (see https://docs.oracle.com/cd/E19509-01/820-3503/ggfhb/index.html) and openssl supports both pkcs12 and jks.  However, I can see you are now only loading certs now. Perhaps this may be sufficient now? (depends on how we buy/generate the certs/keys)

ijmad: Sure. I've tested this as working with self-signed certificates. I don't want this merged quite yet - but it does seem to work.

src/main/java/uk/gov/pay/api/utils/TrustStoreLoader.java
ijmad: Commit is marked as experimental for a reason!

I don't follow what you mean by "remove the certs after a successful load.". The certs may change between container restarts and we need to notice this and pick up the new certs when they change.

ijmad: This should only load when the JVM starts, isn't this what we want? Certs may change, and a restart should reload new certs.

kakumara: I assume this will get called during a start/restart of Dropwizard app. So this will try to reload the certs from `CERTS_PATH` during a restart even if was loaded before. If we remove the files from CERTS_PATH after a successful load we can perhaps avoid it? (unless we can be sure that trust stores handles this magically by not accepting the same cert multiple times)

ijmad: CERTS_PATH might be a mounted docker volume, so removing the certs might be a bad idea. I don't really follow your point though, surely a restart of the DropWizard app is exactly when we want to check for cert changes. The TrustStore is new every time, so there's no chance of overlapping or adding the same cert twice

ijmad:
Summarizing our discussion:
- There is no file on disk to worry about now as this is entirely in memory
- Container/JVM restart is the right time to re-create the truststore as externally provided certs may change at any time and a restart is an easy time to check this

I have removed the docker-startup.sh script that was left over but generating a cacerts file no-longer used. This is now ready for re-review and potential merge.

ijmad:
PR cleaned up and reopened as https://github.com/alphagov/pay-publicapi/pull/75

PR #72, opened 2016-03-31
PP-486 Deserialisation next_url_post from Connector - Merge after pay-connector PP-486

rauligs:
- Created new extra link _next_post_url_ coming from Connector.
- Charges from connector are now deserialised using  Charge class rather than reading Json fields.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/PaymentWithLinks.java
+++ b/src/main/java/uk/gov/pay/api/model/PaymentWithLinks.java
@@ -2,6 +2,7 @@ package uk.gov.pay.api.model;
 
 import com.fasterxml.jackson.databind.JsonNode;
 
+import java.net.URI;
 import java.util.Iterator;
 import java.util.Optional;
 
@@ -14,16 +15,20 @@ public class PaymentWithLinks implements PaymentWithLinksJSON {
         return links;
     }
 
-    public static PaymentWithLinks createPaymentResponseWithLinks(JsonNode payload, String selfLink) {
-        Payment payment = Payment.createPaymentResponse(payload);
-        PaymentWithLinks paymentWithLinks = new PaymentWithLinks(payment);
-        paymentWithLinks.withSelfLink(selfLink);
-        Optional<JsonNode> nextLinkMaybe = getNextLink(payload);
-
-        nextLinkMaybe.ifPresent(
-                (nextLink) -> paymentWithLinks.withNextLink(nextLink.get("href").asText()));
-
-        return paymentWithLinks;
+    public static PaymentWithLinks valueOf(JsonNode payload, URI selfLink) {
+        PaymentWithLinks payment = new PaymentWithLinks(Payment.valueOf(payload, selfLink));
+        payment.withSelfLink(selfLink.toString());
+        JsonNode links = payload.get("links");
+        links.forEach(link -> {
+            String rel = link.get("rel").asText();
+            if ("next_url".equals(rel)) {
+                payment.withNextLink(link.get("href").asText());
+            }
+            if ("next_url_post".equals(rel)) {
+                payment.withNextPostLink(link.get("href").asText(), link.get("type").asText(), link.get("params"));
+            }
+        });
+        return payment;
     }
 
     private PaymentWithLinks(Payment payment) {
@@ -72,15 +77,8 @@ public class PaymentWithLinks implements PaymentWithLinksJSON {
         return this;
     }
 
-    private static Optional<JsonNode> getNextLink(JsonNode payload) {
-        for (Iterator<JsonNode> it = payload.get("links").elements(); it.hasNext(); ) {
-            JsonNode node = it.next();
-            if ("next_url".equals(node.get("rel").asText())) {
-                return Optional.of(node);
-            }
-        }
-
-        return Optional.empty();
+    public PaymentWithLinks withNextPostLink(String url, String type, JsonNode params) {
+        this.links.setNextUrlPost(url, type, params);
+        return this;
     }
-
 }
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/api/model/Charge.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public String getCreated_date() {
+        return created_date;
+    }
+
+    public List<ChargeLink> getLinks() {
+        return links;
+    }
+}
```
imranbohoran: I'm not sure if having another object to represent the concept of 'Charge' from connector is appropriate here. As a model for the public-api, we have Payment as that's the terminology used where as a 'Charge' is more connector lingo. So I wonder if its going to be mis-interpreted and if it would be confusing. 
I do see the benefit of letting jackson parse the object and in terms of something changing in the connector response, changes will be the doing it this way or manually reading them from a JsonNode. 
Perhaps we name it differently as 'ConnectorResponse' as oppose to a charge? I might be over thinking this, but at first glance I got a bit confused knowing there's a Charge concept in connector.

src/main/java/uk/gov/pay/api/model/ChargeLink.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public Map<String, Object> getParams() {
+        return params;
+    }
+}
```
imranbohoran: same comment as Charge above

src/main/java/uk/gov/pay/api/model/PaymentWithLinks.java
```diff
@@ -1,33 +1,21 @@
...
-        return paymentWithLinks;
+    public static PaymentWithLinks valueOf(Charge charge, URI selfLink) {
+        return new PaymentWithLinks(Payment.valueOf(charge), charge.getLinks(), selfLink);
     }
 
-    private PaymentWithLinks(Payment payment) {
+    private PaymentWithLinks(Payment payment, List<ChargeLink> chargeLinks, URI selfLink) {
         this.payment = payment;
+        this.links.addSelf(selfLink.toString());
+        this.links.addKnownLinksValueOf(chargeLinks);
```
imranbohoran: :+1:

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -252,7 +251,7 @@ public Response cancelPayment(@ApiParam(value = "accountId", hidden = true) @Aut
     }
 
     private String errorMessageFrom(List<Pair<String, String>> invalidParams) {
-        List<String> keys = invalidParams.stream().map(pair -> pair.getLeft()).collect(Collectors.toList());
+        List<String> keys = invalidParams.stream().map(Pair::getLeft).collect(Collectors.toList());
```
imranbohoran: :+1:

src/test/java/uk/gov/pay/api/it/PaymentsResourceITest.java
```diff
@@ -183,18 +197,15 @@ public void getPaymentEvents_ReturnsPaymentEvents() {
...
-        assertEquals(list.get(0).get("payment_id"), CHARGE_ID);
-        assertEquals(list.get(0).get("status"), STATUS);
-        assertEquals(list.get(0).get("updated"), "2016-01-01 12:00:00");
-
-        response.body("_links.self.href", is(paymentEventsLocationFor(CHARGE_ID)));
+                .body("events", hasSize(1))
+                .body("events[0].payment_id", is(CHARGE_ID))
+                .body("events[0].status", is(STATUS))
+                .body("events[0].updated", is("2016-01-01 12:00:00"))
+                .body("_links.self.href", is(paymentEventsLocationFor(CHARGE_ID)));
```
imranbohoran: :+1:

src/test/java/uk/gov/pay/api/utils/ConnectorMockClient.java
```diff
@@ -60,11 +63,14 @@ private String createChargeResponse(long amount, String chargeId, String status,
                 .add("return_url", returnUrl)
                 .add("payment_provider", paymentProvider)
                 .add("created_date", createdDate)
-                .add("links", asList(links))
-                .build();
+                .add("links", asList(links));
+        if (gatewayTransactionId != null) {
+            jsonStringBuilder.add("gateway_transaction_id", gatewayTransactionId);
+        }
```
imranbohoran: This is not something we map into the Payment object of the public-api, do we really want to leak connector concerpt of gateway_transaction_id to the public-api?

rauligs: Indeed, you see that because effectively Connector is sending it back but we are ignoring it and It wasn't reflected in the tests so I created a failing test and making sure we don't map this field.

src/main/java/uk/gov/pay/api/model/Charge.java
rauligs: Connector return more things, what about ChargeConnectorResponse? what do you suggest?

imranbohoran: As you suggested PaymentConnectorResponse might be a good option

src/test/java/uk/gov/pay/api/utils/ConnectorMockClient.java
imranbohoran: I guess this is because we are letting jackson map the object to the json (with the IgnoreUnknownProperties). Fair enough having some test case around it as well.

imranbohoran:
rebuild

PR #73, opened 2016-03-31
PP-539 Selectively including links in the payment responses

imranbohoran:
This addresses the need to have separate response structures for
'create', 'find by payment id' and 'search payments'. Providing links in
the response for search payments doesn't make sense as it presents a
snapshot of each payment. However 'create' and 'find by payment id'
require links to be present in their responses.

* Introducing a concept of PaymentWithLinks to represent the fact that
its different from a Payment
* All JSON and API documentation annotations are moved to interfaces to
give the value objects json behaviors.
* The objects would just be value objects with the interfaces giving
them JSON and API documentation capabilities.
* Both create payments and find payment now uses the new PaymentWithLink
object as it makes sense to include links in these responses.
* Search payment will use the Payment object which doesn't include
links.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/PaymentJSON.java
+++ b/src/main/java/uk/gov/pay/api/model/PaymentJSON.java
@@ -1,0 +1,41 @@
+package uk.gov.pay.api.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel(description = "A Payment")
+interface PaymentJSON {
+
+    @JsonProperty("created_date")
+    @ApiModelProperty(example = "2016-01-21T17:15:00Z")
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/api/model/PaymentJSON.java
```diff
@@ -0,0 +1,41 @@
+package uk.gov.pay.api.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel(description = "A Payment")
+interface PaymentJSON {
```
kakumara: Nice :+1:  that we have all the documentation annotations abstracted aways from implementation. (Although `Fran` tried this sometime ago and found out some complications in swagger. Not sure why though, but if this works then good)

kakumara:
Looks good to me :+1:

PR #74, opened 2016-03-31
PP-539 Making the create payment payload restrictive.

imranbohoran:
* Removing the ignore unknown properties capability when parsing a json
payload to create a payment.
* If the payload has an attribute that we don't know about, the request
will response with a 400.


kakumara:
:+1:

PR #75, opened 2016-04-04
PP 618 fix cert loading

ijmad:
There is no need for the private keys.
cacerts is a truststore, not a keystore.

Move to a programmatic approach for creating a truststore from the pem directory, to avoid the need for bash scripting inside the containers.

### DIFF
--- a/src/main/java/uk/gov/pay/api/validation/URLValidator.java
+++ b/src/main/java/uk/gov/pay/api/validation/URLValidator.java
@@ -6,7 +6,7 @@ import static org.apache.commons.validator.routines.UrlValidator.*;
 
 public enum URLValidator {
     SECURITY_ENABLED {
-        private final UrlValidator URL_VALIDATOR = new UrlValidator(new String[]{"https"});
+        private final UrlValidator URL_VALIDATOR = new UrlValidator(new String[]{"https"}, ALLOW_LOCAL_URLS);
 
         @Override
         public boolean isValid(String value) {
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ hs_err_pid*
 target
 
 bin
+
+.DS_Store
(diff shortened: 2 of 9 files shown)


swagger/swagger.json
```diff
@@ -302,6 +297,11 @@
           "type" : "string",
           "example" : "worldpay",
           "readOnly" : true
+        },
+        "status" : {
+          "type" : "string",
+          "example" : "CREATED",
+          "readOnly" : true
```
ijmad: I'm not really sure why this change has moved this attribute in swagger, but it seems harmless enough.

swagger/swagger.json
```diff
@@ -267,29 +267,29 @@
           "example" : "your-reference",
           "readOnly" : true
         },
-        "created_date" : {
+        "description" : {
```
kakumara: I cannot see you've done any changes to Documentation/Examples. Are these intentional changes?

src/main/java/uk/gov/pay/api/validation/URLValidator.java
```diff
@@ -6,7 +6,7 @@
 
 public enum URLValidator {
     SECURITY_ENABLED {
-        private final UrlValidator URL_VALIDATOR = new UrlValidator(new String[]{"https"});
+        private final UrlValidator URL_VALIDATOR = new UrlValidator(new String[]{"https"}, ALLOW_LOCAL_URLS);
```
kakumara: Is this a good thing? This means that If someone sends us a `localhost`  `return_url` we accept it?

ijmad: Yeah, it's a bit controversial, but people will be coming from localhost for sandbox testing. Maybe we need some more discussion. I'll withdraw this change for now.

kakumara: 👍

PR #77, opened 2016-04-04
PP-539 Allowing self links in payment search

imranbohoran:
PP-539 Allowing self link for payments from search    
     * Including self link in payment search response. Only self link is
       available in the '_links' object of the payment search response.
     * Introducing separate objects to represent search results, links that
       only have safe links and links that have both safe links and next_url
       links.
     * Every response other than payment search and cancel will have the
       next_url in the links object
     * Adding tests to assert we are returning everything we need to return
     * Fixing typo in documentation

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/PaymentWithSelfLinksJSON.java
+++ b/src/main/java/uk/gov/pay/api/model/PaymentWithSelfLinksJSON.java
@@ -4,13 +4,13 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 
-@ApiModel(description = "A Payment with self and next links")
-public interface PaymentWithLinksJSON {
+@ApiModel(description = "A Payment with only a self link")
+public interface PaymentWithSelfLinksJSON {
 
     String LINKS_JSON_ATTRIBUTE = "_links";
 
     @JsonProperty(LINKS_JSON_ATTRIBUTE)
-    @ApiModelProperty(dataType = "uk.gov.pay.api.model.SelfAndNextLinks")
-    SelfAndNextLinks getLinks();
+    @ApiModelProperty(dataType = "uk.gov.pay.api.model.SelfLinks")
+    SelfLinks getLinks();
 
 }
--- a/src/main/java/uk/gov/pay/api/model/SelfAndNextLinks.java
+++ b/src/main/java/uk/gov/pay/api/model/SelfAndNextLinks.java
@@ -1,0 +1,72 @@
+package uk.gov.pay.api.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import java.util.List;
+
+import static javax.ws.rs.HttpMethod.GET;
+import static uk.gov.pay.api.model.SelfLinks.SELF;
+
+@ApiModel(value = "selfAndNextLinks", description = "Resource self and next links of a Payment")
+public class SelfAndNextLinks {
+
+    private static final String NEXT_URL = "next_url";
+    private static final String NEXT_URL_POST = "next_url_post";
+
+    private Link self;
+
+    @JsonProperty(NEXT_URL)
+    private Link nextUrl;
+
+    @JsonProperty(NEXT_URL_POST)
+    private Link nextUrlPost;
+
+    @ApiModelProperty(value = SELF, dataType = "uk.gov.pay.api.model.Link")
+    public Link getSelf() {
+        return self;
+    }
+
+    @ApiModelProperty(value = NEXT_URL, dataType = "uk.gov.pay.api.model.Link")
+    public Link getNextUrl() {
+        return nextUrl;
+    }
+
+    @ApiModelProperty(value = NEXT_URL_POST, dataType = "uk.gov.pay.api.model.Link")
+    public Link getNextUrlPost() {
+        return nextUrlPost;
+    }
+
+    @Override
+    public String toString() {
+        return "Links{" +
+                "self=" + self +
+                ", nextUrl=" + nextUrl +
+                ", nextUrlPost=" + nextUrlPost +
+                '}';
+    }
+
+    void addSelf(String href) {
+        this.self = new Link(href, GET);
+    }
+
+    void addKnownLinksValueOf(List<PaymentConnectorResponseLink> chargeLinks) {
+        addNextUrlIfPresent(chargeLinks);
+        addNextUrlPostIfPresent(chargeLinks);
+    }
+
+    private void addNextUrlPostIfPresent(List<PaymentConnectorResponseLink> chargeLinks) {
+        chargeLinks.stream()
+                .filter(chargeLink -> NEXT_URL_POST.equals(chargeLink.getRel()))
+                .findFirst()
+                .ifPresent(chargeLink -> this.nextUrlPost = new Link(chargeLink.getHref(), chargeLink.getMethod(), chargeLink.getType(), chargeLink.getParams()));
+    }
+
+    private void addNextUrlIfPresent(List<PaymentConnectorResponseLink> chargeLinks) {
+        chargeLinks.stream()
+                .filter(chargeLink -> NEXT_URL.equals(chargeLink.getRel()))
+                .findFirst()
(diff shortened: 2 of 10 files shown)


imranbohoran:
rebuild

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -142,7 +151,7 @@ public Response searchPayments(@ApiParam(value = "accountId", hidden = true)
                                    @Auth String accountId,
                                    @ApiParam(value = "Your payment reference to search", hidden = false)
                                    @QueryParam(REFERENCE_KEY) String reference,
-                                   @ApiParam(value = "Status of payments to be searched. Example=SUCCESSED", hidden = false, allowableValues = "range[SUCCEEDED,CREATED,IN PROGRESS,FAILED,SYSTEM CANCELLED")
+                                   @ApiParam(value = "Status of payments to be searched. Example=SUCCEEDED", hidden = false, allowableValues = "range[SUCCEEDED,CREATED,IN PROGRESS,FAILED,SYSTEM CANCELLED")
```

src/main/java/uk/gov/pay/api/model/PaymentWithLinks.java
```diff
@@ -3,10 +3,10 @@
 import java.net.URI;
 import java.util.List;
 
-public class PaymentWithLinks implements PaymentWithLinksJSON {
+public class PaymentWithLinks implements PaymentWithLinksJSON, PaymentJSON {
```
PauloPortugal: Although we are following the _Open/Close_ and _Single Responsibility_ principles here and using a _Decorator pattern_ to extend the `Payment` object, I believe this is a bit over engineered as we now have `PaymentWithLinks` and `PaymentWithSelfLinks` which is with a bit of hindsight are terrible names (although not as bad as celebrity children names), because any other developer reading this code in the future will keep on asking why do we have these names and we are raising our tech debt.

If every payment will have a `_self` link, then I would made the _links with `_self` as part of the `Payment` object with a generic `Links` attribute and that would make use inheritance if we want to use the NextLinks or any other links.

src/main/java/uk/gov/pay/api/model/PaymentWithSelfLinksJSON.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.pay.api.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel(description = "A Payment with only a self link")
+public interface PaymentWithSelfLinksJSON {
```
PauloPortugal: This is the same as PaymentWithLinksJSON with the exception of the `@ApiModelProperty`.
We can sit down and find a better name convention or implementation..

src/main/java/uk/gov/pay/api/model/SelfAndNextLinks.java
```diff
@@ -0,0 +1,72 @@
...
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import java.util.List;
+
+import static javax.ws.rs.HttpMethod.GET;
+import static uk.gov.pay.api.model.SelfLinks.SELF;
+
+@ApiModel(value = "selfAndNextLinks", description = "Resource self and next links of a Payment")
+public class SelfAndNextLinks {
```
PauloPortugal: This is a case where we could make use of inheritance and then simply annotate the methods required for documentation.
This will move the documentation annotation into the concrete classes, but I think it is a lesser evil than having all this boiler plate code for documentation sake.

src/main/java/uk/gov/pay/api/model/SelfLinks.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.pay.api.model;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import static javax.ws.rs.HttpMethod.GET;
+
+@ApiModel(value = "selfLinks", description = "Resource self link of a Payment")
+public class SelfLinks {
```
PauloPortugal: I would rename this to simply `Links` and anything more complex could you the _Decorator pattern_ and we could call it `ExtendedLinks`, instead of calling `SelfAndNextLinks`, as any other change of implementation would require a change of name as well (again we are being a bit more loose with _Cohesion_ and _Single Responsibility_ principles) but covering our selves for future tech debt and I believe better code readability.

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -134,7 +143,7 @@ public Response getPaymentEvents(@ApiParam(value = "accountId", hidden = true) @
             responseContainer = "List",
             code = 200)
 
-    @ApiResponses(value = {@ApiResponse(code = 200, message = "OK", response = Payment.class, responseContainer = "List"),
+    @ApiResponses(value = {@ApiResponse(code = 200, message = "OK", response = PaymentSearchResults.class),
```

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -286,6 +297,41 @@ private Response responseForPaymentWithLinks(UriInfo uriInfo, Response connector
...
 
+    private Response responseForPaymentsWithSelfLinks(UriInfo uriInfo, Response connectorResponse, int okStatus,
+                                                      Function<Object, ResponseBuilder> okResponse,
+                                                      Supplier<Response> errorResponse) {
+
+        if (connectorResponse.getStatus() == okStatus) {
+            try {
+                JsonNode responseJson = connectorResponse.readEntity(JsonNode.class);
+
+                TypeReference<HashMap<String, List<PaymentConnectorResponse>>> typeRef
```

src/main/java/uk/gov/pay/api/model/PaymentWithLinks.java
imranbohoran: I do agree the interfaces would be unnecessary given the current link structure we've agreed to. I've actually started working on getting rid of them in the next PR i'm creating (that fixes the PaymentEvents).
I was thinking in the lines of having a Payment object which has the API* annotations and Jackson annotations. 
But the links will be in their own classes. We can't have a generic '_links' attribute in the Payment class since Jackson and swagger needs the class structure to be in place in the class. In the case of having a '_self' and '_next_link', we have to have a object that is annotated with '_links' and we need to have '_self' and '_next_link' annotated inside that class. Else we don't get the same json structure we expect.
So essentially we'd have a hierarchy of Payment->PaymentWithSelfLink and Payment->PaymentWithSelfAndNextLink

** Check the comment below about naming

src/main/java/uk/gov/pay/api/model/PaymentWithSelfLinksJSON.java
imranbohoran: I think I've covered this in the above comment.

src/main/java/uk/gov/pay/api/model/SelfAndNextLinks.java
imranbohoran: Agreed. Please see the reply to the first comment.

src/main/java/uk/gov/pay/api/model/SelfLinks.java
imranbohoran: I thought the class described what its suppose to do if we name it as 'SelfLinks' and 'SelfAndNextLinks'. I've been able to use the same model when fixing up the PaymentEvents as well. I do agree _if_ we have to add a new link to the '_link' structure then we have to either add a new one or amend the name. But as of now I'm not aware of any reason to do this, other than at some point we'd want to remove '_self' from '_links' and make it its own object. In that event we can rename the 'SlefAndNextLinks' to 'Links', remove 'SelfLinks' and move the whole '_self' object into payments. But there's no assurances _if/when_ that would happen. So I thought of making the naming very specific - just following YAGNI.
If you still feel strongly about the naming I'm happy to change it as well though.

Another thing worth noting is that 'Link' class currently exists in this PR for the purpose of PaymentEvents. This is going away when PaymentEvents are fixed as it has its own 'PaymentEventLink' object. See https://payments-platform.atlassian.net/browse/PP-539 for details.

PauloPortugal:
I have spoke with with @imranbohoran about the solution and the problem stems from Swagger and Jackson that require the class structure to be present.
@imranbohoran also agrees that the `Payment` structure requires some refactoring and we are happy for that work to be done as part of the Payment Events PR.

PR #78, opened 2016-04-06
PP-539 Fixing payment events responses

imranbohoran:
- Fixes the payment events responses to adhere to the correct structure. It includes an outer `self_ url` to the payment events resource and a `payment_url` in the individual payment event objects.
- Getting rid of the interfaces and introducing a class hierarchy for payments.
- Any `_link` at the moment has `params` and `type` included. I don't think we need this for everything other than the POST `next_url`s. Will talk to @rauligs to see if we can change so links that need post urls have their own class.


src/main/java/uk/gov/pay/api/model/Payment.java
```diff
@@ -1,33 +1,37 @@
...
 
-public class Payment implements PaymentJSON {
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModelProperty;
+
+public abstract class Payment {
+
+    public static final String LINKS_JSON_ATTRIBUTE = "_links";
+
+    @JsonProperty("payment_id")
```
PauloPortugal: :+1:

src/main/java/uk/gov/pay/api/model/PaymentEvent.java
```diff
@@ -13,10 +13,12 @@
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 
-@ApiModel(value="Payment Events information", description = "A List of Payment Events information")
+@ApiModel(value="Payment Event information", description = "A List of Payment Events information")
 public class PaymentEvent {
     @JsonProperty("payment_id")
     private final String paymentId;
+
+    @JsonProperty("status")
```
PauloPortugal: Is this really needed?

src/main/java/uk/gov/pay/api/model/PaymentEvent.java
```diff
@@ -24,23 +26,28 @@
     @JsonFormat(pattern="yyyy-MM-dd HH:mm:ss")
     private final LocalDateTime updated;
 
-    public static PaymentEvent createPaymentEvent(JsonNode payload) {
+    @JsonProperty("_links")
```
PauloPortugal: :+1:

src/main/java/uk/gov/pay/api/model/Payment.java
PauloPortugal: I am not a big fan of using JsonProperties on attributes that do not need them, but I understand that we might want them there for code consistency.

src/main/java/uk/gov/pay/api/model/PaymentEventLink.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.pay.api.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import static javax.ws.rs.HttpMethod.GET;
+
+@ApiModel(value = "payment-event-link", description = "Resource link for a payment of a payment event")
+public class PaymentEventLink {
```
PauloPortugal: :+1:

src/main/java/uk/gov/pay/api/model/PaymentEvent.java
imranbohoran: It would pick it up without it. I preferred being explicit on the json names. It doesn't add any overhead, but added for consistency.

src/main/java/uk/gov/pay/api/model/PaymentWithLinks.java
```diff
@@ -1,55 +1,42 @@
 package uk.gov.pay.api.model;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModelProperty;
+
 import java.net.URI;
 import java.util.List;
 
-public class PaymentWithLinks implements PaymentWithLinksJSON, PaymentJSON {
+public class PaymentWithLinks extends Payment {
```
PauloPortugal: :+1:

PR #79, opened 2016-04-06
PP-611 500 Error mapping Creating a Charge when Gateway account does not exist

rauligs:
- When Authorization succeed but Gateway account is Not Found in Connector, PublicAPI returns a 500 rather than a 400 Bad Request. This situation can't theoretically happen and a client can't recover from this without support.

### DIFF
--- a/src/main/java/uk/gov/pay/api/exception/CreateChargeConnectorErrorResponseExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/api/exception/CreateChargeConnectorErrorResponseExceptionMapper.java
@@ -2,11 +2,15 @@ package uk.gov.pay.api.exception;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.api.model.PaymentErrorResponse;
 
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 
-import static javax.ws.rs.core.Response.Status.*;
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static uk.gov.pay.api.exception.CreateChargeConnectorErrorResponseExceptionMapper.CreateChargeErrorResponse.P0198;
+import static uk.gov.pay.api.exception.CreateChargeConnectorErrorResponseExceptionMapper.CreateChargeErrorResponse.P0199;
 
 public class CreateChargeConnectorErrorResponseExceptionMapper implements ExceptionMapper<CreateChargeConnectorErrorResponseException> {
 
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/api/exception/CreateChargeConnectorErrorResponseExceptionMapper.java
```diff
@@ -0,0 +1,53 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(CreateChargeConnectorErrorResponseException.class);
+
+    @Override
+    public Response toResponse(CreateChargeConnectorErrorResponseException exception) {
+
+        int errorStatus = exception.getErrorStatus();
+
+        if (errorStatus == NOT_FOUND.getStatusCode()) {
+            LOGGER.warn("Authorization succeeded but Gateway account was not found in Connector", errorStatus);
+            return P0199.response();
```
rahul-vinayak: is logger.warn good enough for this, I see this as a rare and severe exception to deal with, should be error, or even fatal

src/main/java/uk/gov/pay/api/exception/CreateChargeConnectorErrorResponseExceptionMapper.java
```diff
@@ -0,0 +1,53 @@
...
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static uk.gov.pay.api.exception.CreateChargeConnectorErrorResponseExceptionMapper.CreateChargeErrorResponse.P0198;
+import static uk.gov.pay.api.exception.CreateChargeConnectorErrorResponseExceptionMapper.CreateChargeErrorResponse.P0199;
+
+public class CreateChargeConnectorErrorResponseExceptionMapper implements ExceptionMapper<CreateChargeConnectorErrorResponseException> {
+
```
rahul-vinayak: is it possible to make this mapper class and the exception class more generic one, or we will end having very specific exception classes and the mappers. Also making it more generic will move away from such a long class names

src/main/java/uk/gov/pay/api/exception/CreateChargeConnectorErrorResponseExceptionMapper.java
```diff
@@ -0,0 +1,53 @@
...
+        LOGGER.info("Error response from Connector unrecognised. status {}, body {}", errorStatus, exception.getErrorBody());
+        return P0198.response();
+    }
+
+    enum CreateChargeErrorResponse {
+
+        P0199(INTERNAL_SERVER_ERROR, "There is an error with this account. Please contact support"),
+        P0198(INTERNAL_SERVER_ERROR, "Downstream system error");
+
```
rahul-vinayak: 🎱 👍

src/main/java/uk/gov/pay/api/exception/CreateChargeConnectorErrorResponseExceptionMapper.java
rauligs: Will change it to error

src/main/java/uk/gov/pay/api/exception/CreateChargeConnectorErrorResponseExceptionMapper.java
rauligs: This story is in progress, If during the refactoring process makes sense will be changed to something more generic but not as part of this PR.

PR #81, opened 2016-04-08
PP-539 Fixing payment event dates and Links

imranbohoran:
- The payment event dates were using LocalDateTime. This has been changed now to use what the connector would return (as we do with dates returned by other end-points). Connector returns ZoneDateTime.
- The next_url_post has type and params as documentation. These don't apply to GET links. Therefore GET links and POST links are separated.
- Making it clear that authroization header requires Bearer <TOKEN> as swagger fails to document this nicely.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/PaymentEvent.java
+++ b/src/main/java/uk/gov/pay/api/model/PaymentEvent.java
@@ -1,17 +1,9 @@
 package uk.gov.pay.api.model;
 
-import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import com.fasterxml.jackson.datatype.joda.deser.LocalDateTimeDeserializer;
-import com.fasterxml.jackson.datatype.jsr310.ser.LocalDateTimeSerializer;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
-
-import java.time.LocalDateTime;
-import java.time.format.DateTimeFormatter;
 
 @ApiModel(value="Payment Event information", description = "A List of Payment Events information")
 public class PaymentEvent {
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/api/model/PaymentEvent.java
```diff
@@ -21,29 +13,22 @@
     @JsonProperty("status")
     private final String status;
 
-    @JsonDeserialize(using = LocalDateTimeDeserializer.class)
-    @JsonSerialize(using = LocalDateTimeSerializer.class)
-    @JsonFormat(pattern="yyyy-MM-dd HH:mm:ss")
-    private final LocalDateTime updated;
+    @JsonProperty("updated")
```
PauloPortugal: :+1:

src/main/java/uk/gov/pay/api/model/PaymentEvent.java
```diff
@@ -21,29 +13,22 @@
...
-        LocalDateTime updated = null;
-        if(payload.get("updated") != null) {
-            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
-            updated = LocalDateTime.parse(payload.get("updated").asText(), formatter);
-        }
         return new PaymentEvent(
             payload.get("charge_id").asText(),
             payload.get("status").asText(),
-            updated,
+            payload.get("updated").asText(),
```
PauloPortugal: :+1:

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -87,7 +87,9 @@ public PaymentsResource(Client client, String connectorUrl) {
     @Produces(APPLICATION_JSON)
     @ApiOperation(
             value = "Find payment by ID",
-            notes = "Return information about the payment",
+            notes = "Return information about the payment " +
```

src/test/java/uk/gov/pay/api/it/PaymentSearchITest.java
```diff
@@ -2,6 +2,7 @@
 
 
 import com.google.common.collect.ImmutableMap;
+import com.jayway.jsonassert.JsonAssert;
```
PauloPortugal: :+1:

PauloPortugal:
:+1:

PR #82, opened 2016-04-12
PP-640 Add events link in payment objects

imranbohoran:
**WHAT**
Introducing event links in payment objects. Every payment object will have a events link in the '_link' object. The only exception is the GET on payment events. As the commits explain there has been some restructuring and renaming of classes. Please also note that there's another part to the story where cancel links need to be added to payment objects.

**WHO**
any developer


imranbohoran:
rebuild

swagger/swagger.json
```diff
@@ -435,37 +488,6 @@
...
-        "self" : {
-          "description" : "self",
-          "readOnly" : true,
-          "$ref" : "#/definitions/paymentLink"
-        }
-      },
-      "description" : "Resource self link of a Payment"
     }
   }
 }
```
rauligs: Configure to add new lines automatically

rauligs:
Get payment - check it does not contain events link

swagger/swagger.json
imranbohoran: I checked and I have it setup in intelliJ. But given that this file is generated by the maven plugin and we don't hand-craft it this is what we get

imranbohoran:
Re: test for GET payment. Thanks for that, it's added now.

PR #86, opened 2016-04-18
PP-611 Normalize Search Payments API

rauligs:
- Extracted validations and reuse existing ValidationExceptions.
    - Added validations for reference and description maxLength.
    - Logging error codes.


rauligs:
rebuild

src/main/java/uk/gov/pay/api/model/PaymentError.java
```diff
@@ -15,7 +15,9 @@
         P0100,
         P0101,
         P0102,
-        P0103
+        P0103,
+        P0401,
+        P0498
```
imranbohoran: This is a general comment about the PaymentError class. Do you think it makes sense to model this in a way where we capture a description of the Code within the Code enum. Looking at the Code enum the code doesn't give out any description unless you look at the different static methods. But that too the description is an argument.

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -234,12 +231,12 @@ public Response searchPayments(@ApiParam(value = "accountId", hidden = true)
 
                 return Response.ok(new PaymentSearchResults(paymentsForSearchResults)).build();
 
-            } catch (IOException e) {
-                return serverErrorResponse(logger, "Search payments failed");
+            } catch (IOException | ProcessingException e) {
+                throw new SearchChargesException(e);
```
imranbohoran: 👍

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -285,7 +282,7 @@ public Response createNewPayment(@ApiParam(value = "accountId", hidden = true) @
             return Response.created(paymentUri).entity(payment).build();
 
         } else {
-            throw new CreateChargeConnectorErrorResponseException(connectorResponse.getStatus(), connectorResponse.readEntity(String.class));
+            throw new CreateChargeException(connectorResponse);
         }
     }
```
imranbohoran: 👍 For the exceptions

src/test/java/uk/gov/pay/api/it/PaymentResourceSearchITest.java
```diff
@@ -7,34 +7,45 @@
...
     private static final String TEST_STATUS = "created";
     private static final String TEST_FROM_DATE = "2016-01-28T00:00:00Z";
     private static final String TEST_TO_DATE = "2016-01-28T12:00:00Z";
     private static final String SEARCH_PATH = "/v1/payments";
 
+    @Before
+    public void mapBearerTokenToAccountId() {
+        publicAuthMock.mapBearerTokenToAccountId(BEARER_TOKEN, GATEWAY_ACCOUNT_ID);
+    }
+
```
imranbohoran: 👍

src/main/java/uk/gov/pay/api/model/PaymentError.java
rauligs: When the last pull request is done will be another one to try to do what do you say (completely agree)

rauligs:
rebuild

src/main/java/uk/gov/pay/api/model/PaymentError.java
imranbohoran: 👍

PR #87, opened 2016-04-19
PP-611 Normalize Get Payment, Events and Cancel API resources

rauligs:
- Created exceptions for invalid responses.
    - Make sure for every one readEntity is in place.
    - Close connections when entity has not been read.
    - Delegate error construction an formatting to PaymentError so this way we have all the description/codes in one place.


rauligs:
rebuild

src/main/java/uk/gov/pay/api/exception/ConnectorResponseErrorException.java
```diff
@@ -4,23 +4,19 @@
...
 class ConnectorResponseErrorException extends RuntimeException {
 
-    private int errorStatus;
-    private String errorBody;
+    private int status;
 
     ConnectorResponseErrorException(Response response) {
-        this.errorStatus = response.getStatus();
-        this.errorBody = response.readEntity(String.class);
+        super(response.toString());
```
kakumara: `toString()` -> Isn't this just the ObjectID of Response class? And don't we always send a error body from connector ? (so that we could readEntity())

rauligs: No it is not, is the response displayed in a nice way (better than read Entity, already checked)

src/main/java/uk/gov/pay/api/model/PaymentError.java
```diff
@@ -1,64 +1,85 @@
...
-        P0101,
-        P0102,
-        P0103,
-        P0401,
-        P0498
-    }
 
-    private final Code code;
-    private final String description;
+        CREATE_PAYMENT_ACCOUNT_ERROR("P0199", "There is an error with this account. Please contact support"),
```
kakumara: 👍  nice

kakumara:
README needs updating perhaps?

src/test/java/uk/gov/pay/api/it/PaymentsResourceDescriptionValidationITest.java
```diff
@@ -23,7 +23,7 @@ public void setUpBearerToken() {
     }
 
     @Test
-    public void createPayment_responseWith400_whenReferenceIsNumeric() throws IOException {
+    public void createPayment_responseWith400_whenDescriptionIsNumeric() throws IOException {
```
kakumara: What's wrong with description being numeric?

src/test/java/uk/gov/pay/api/it/PaymentsResourceITest.java
```diff
@@ -239,16 +238,41 @@ public void getPayment_Returns401_WhenUnauthorised() {
...
+
+        InputStream body = getPaymentResponse(BEARER_TOKEN, paymentId)
+                .statusCode(500)
+                .contentType(JSON).extract()
+                .body().asInputStream();
+
+        JsonAssert.with(body)
+                .assertThat("$.*", hasSize(2))
+                .assertThat("$.code", is("P0298"))
+                .assertThat("$.description", is("Downstream system error"));
```
kakumara: I know you are vetting this with a designer/UR community. But exposing "Down stream error" to outside world doesn't sound like a good thing? Perhaps something like "Problem with payment platform" ?

src/test/java/uk/gov/pay/api/it/PaymentsResourceDescriptionValidationITest.java
rauligs: Numeric as in JSON field. Numbers inside a String JSON field are valid.

src/test/java/uk/gov/pay/api/it/PaymentsResourceITest.java
rauligs: The message is something still to show to them. Before was responding with whatever response came from Connector (and we don't want that)

PR #89, opened 2016-04-25
PP-677 stopped publicapi events end point from throwing null pointer for missing chargeID

rahul-vinayak:
Connector doesn’t return chargeID with every event structure anymore as its in the top level of the structure, which was causing null pointer exception as it was being probed from within the events, this fix stops the chargeId from being probed within every event and looks for it at the top level instead

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/PaymentEvents.java
+++ b/src/main/java/uk/gov/pay/api/model/PaymentEvents.java
@@ -13,6 +13,7 @@ import static uk.gov.pay.api.model.PaymentEvent.createPaymentEvent;
 
 @ApiModel(value="PaymentEventsInformation", description = "A List of Payment Events information")
 public class PaymentEvents {
+    public static final String EVENTS = "events";
     @JsonProperty("payment_id")
     private final String paymentId;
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/api/model/PaymentEvents.java
```diff
@@ -23,9 +24,11 @@
 
     public static PaymentEvents createPaymentEventsResponse(JsonNode payload, String paymentLink) {
         List<PaymentEvent> events = newArrayList();
-        if(payload.get("events").isArray()) {
-            for (JsonNode event : payload.get("events")) {
-                events.add(createPaymentEvent(event, paymentLink));
+        String paymentId = payload.get("charge_id").asText();
```
imranbohoran: Wouldn't this still throw a NPE if there's no `charge_id` in the response from the connector?

Also is it legal to send back some events without paymentIds?

How about some unit tests to cover this?

rahul-vinayak: As clarified (and updated in the PR commit message to be more clear) the chargeID remains at the top level as being returned from the connector, and is not there any longer in every event structure, so there won't a NPE because its looking for chargeID at the right place now. This fact has been covered by the failing E2E test first approach.

Events are not sent back without a paymentID

Will be revisiting the unit tests

rahul-vinayak: Fixed unit test to mock connector to return events as per above description

imranbohoran:
👍

PR #90, opened 2016-04-29
PP-656 Improve healthchecks so they are consistent and more useful

PauloPortugal:
## WHAT
 * The DropWizard healthcheck endpoint is on the admin port and this
   creates much more complexity when routing ports through nginxes, proxies
   and security groups.

## HOW 
 * The Healthcheck should be available on the application port, making use
   of the existing healthchecks that follow Dropwizard's standards.

## WHO
 * Developers and DevOps

### DIFF
--- a/src/main/java/uk/gov/pay/api/resources/HealthCheckResource.java
+++ b/src/main/java/uk/gov/pay/api/resources/HealthCheckResource.java
@@ -1,0 +1,56 @@
+package uk.gov.pay.api.resources;
+
+import com.codahale.metrics.health.HealthCheck;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.collect.ImmutableMap;
+import io.dropwizard.setup.Environment;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Response;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.SortedMap;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.Response.status;
+
+@Path("/")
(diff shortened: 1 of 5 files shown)


imranbohoran:
As discussed, in addition to returning a 200 when everythings running for a node, we should also return a json response with what is failing if the response is a 500.

src/main/java/uk/gov/pay/api/resources/HealthCheckResource.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @GET
+    @Path(HEALTHCHECK)
+    @Produces(APPLICATION_JSON)
+    public Response captureCharge(@PathParam("chargeId") String chargeId) {
+        healthy = environment.healthChecks().getNames().size() == environment.healthChecks().runHealthChecks()
+                .values()
+                .stream()
+                .filter(healthcheck -> healthcheck.isHealthy() == true)
```
imranbohoran: we could possibly simplify this by using a method reference like .filter(HealthCheck.Result::isHealthy)

PauloPortugal:
@imranbohoran I have addressed the 200 vs 500 response.

src/main/java/uk/gov/pay/api/resources/HealthCheckResource.java
```diff
@@ -0,0 +1,56 @@
...
+                .filter(HealthCheck.Result::isHealthy)
+                .count();
+
+        if(healthy) {
+            return Response.ok().entity(response).build();
+        }
+        return status(503).entity(response).build();
+    }
+
+    private Map<String, Map<String, Boolean>> getResponse(SortedMap<String, HealthCheck.Result> results) {
```
imranbohoran: Nice 👍
Can we have a test that ensures we return what we want?

imranbohoran:
👍

PR #91, opened 2016-05-09
PP-541 Pass new state structure through from connector

ijmad:
Depends on https://github.com/alphagov/pay-connector/pull/157 and https://github.com/alphagov/pay-endtoend/pull/166

Pass new state structure through from connector filter params to enable search by the new states. As a result, use the new 'finished' boolean to decide whether to return a cancellation link.

Status remains for now, for backwards compatibility reasons. Removed in next PR.


kakumara:
Few comments in  alphagov/pay-connector#157 which could affect this.  Other than that need this to be green

ijmad:
@kakumara It will only be green once the changes hit connector master

ijmad:
rebuild

rauligs:
rebuild

PR #92, opened 2016-05-09
PP-571 Authorization check - Rate Limiter - depends on publicauth

rauligs:
# WHAT 
So far, Authorization headers were non validated values, provoking for every single request, no matter the Authorization header value (as long it had one), a query into the database.
-  Added an Authorization filter that checks the Authorization header format and also validates that the token is valid by re-creating its Hash value and check the provided Hash being _API Key = Token + Hash_
- Added a Rate Limiter filter after the Authorization validation by limiting the number of requests per a period of time (configurable), current default is 3 requests per second per Authorization (hence a client service).
# WHO
Any dev

### DIFF
--- a/src/main/java/uk/gov/pay/api/app/PublicApi.java
+++ b/src/main/java/uk/gov/pay/api/app/PublicApi.java
@@ -13,6 +13,8 @@ import io.dropwizard.setup.Environment;
 import uk.gov.pay.api.app.config.PublicApiConfig;
 import uk.gov.pay.api.auth.AccountAuthenticator;
 import uk.gov.pay.api.exception.mapper.*;
+import uk.gov.pay.api.filter.RateLimiter;
+import uk.gov.pay.api.filter.RateLimiterFilter;
 import uk.gov.pay.api.healthcheck.Ping;
 import uk.gov.pay.api.json.CreatePaymentRequestDeserializer;
 import uk.gov.pay.api.model.CreatePaymentRequest;
@@ -21,8 +23,10 @@ import uk.gov.pay.api.resources.PaymentsResource;
 import uk.gov.pay.api.validation.PaymentRequestValidator;
 import uk.gov.pay.api.validation.URLValidator;
 
+import javax.servlet.DispatcherType;
 import javax.ws.rs.client.Client;
 
+import static java.util.EnumSet.of;
 import static uk.gov.pay.api.validation.URLValidator.urlValidatorValueOf;
 
 public class PublicApi extends Application<PublicApiConfig> {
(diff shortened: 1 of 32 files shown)


src/main/java/uk/gov/pay/api/app/PublicApi.java
```diff
@@ -41,13 +46,24 @@ public void initialize(Bootstrap<PublicApiConfig> bootstrap) {
...
         environment.jersey().register(new PaymentsResource(client, config.getConnectorUrl()));
-        environment.jersey().register(new HealthCheckResource(environment));
+
+        RateLimiter rateLimiter = new RateLimiter(config.getRateLimiterConfig().getRate(), config.getRateLimiterConfig().getPerMillis());
+
+        environment.servlets().addFilter("AuthorizationValidationFilter", new AuthorizationValidationFilter(config.getApiKeyHmacSecret()))
+                .addMappingForUrlPatterns(of(REQUEST), true, "/v1/*");
+
+        environment.servlets().addFilter("RateLimiterFilter", new RateLimiterFilter(rateLimiter, objectMapper))
+                .addMappingForUrlPatterns(of(REQUEST), true, "/v1/*");
```
kakumara: Did you raise a JIRA to fix the URL mappings properly? (to get rid of this 'v1')

src/main/java/uk/gov/pay/api/app/config/RateLimiterConfig.java
```diff
@@ -0,0 +1,28 @@
...
+import io.dropwizard.Configuration;
+
+import javax.validation.Valid;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+
+public class RateLimiterConfig extends Configuration {
+
+    @Valid
+    @NotNull
```
kakumara: `NotNull` with a primitive variable? And Min/Max any reason for limiting to these numbers ?

src/main/java/uk/gov/pay/api/app/PublicApi.java
rauligs: [PP-693](https://payments-platform.atlassian.net/browse/PP-693)
Have a look to check if the explanation is clear enough.

src/main/java/uk/gov/pay/api/app/config/RateLimiterConfig.java
rauligs: 👍  It was indeed wrong. Fixed validations and removed @NotNull over primitives.

mrlumbu:
rebuild

rauligs:
rebuild

PR #93, opened 2016-05-12
PP-571 Relax rate limiter default values

rauligs:
# WHAT    
- Set default Rate limiter to 25/second for now since the current 3/second is very strict and there is no need.

# WHO
- Any dev/webops


rauligs:
rebuild

imranbohoran:
👍

PR #95, opened 2016-05-12
PP-541 Remove old status field.

ijmad:
Now we have the 'state' field everywhere, we can remove
the old legacy status from the publicapi.

Improve swagger here and there.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/PaymentState.java
+++ b/src/main/java/uk/gov/pay/api/model/PaymentState.java
@@ -4,10 +4,12 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
+import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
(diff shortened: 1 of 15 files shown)


ijmad:
rebuild

ijmad:
rebuild

src/main/java/uk/gov/pay/api/model/PaymentState.java
```diff
@@ -4,20 +4,25 @@
...
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
+import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
+@ApiModel(value="Payment state", description = "A structure representing the current state of the payment in its lifecycle.")
 public class PaymentState {
     @JsonProperty("status")
```
kakumara: Shall we give a swagger example for status too?

src/main/java/uk/gov/pay/api/model/PaymentState.java
```diff
@@ -4,20 +4,25 @@
...
     private String status;
 
+    @ApiModelProperty(value = "Whether the payment has finished", required = true)
```
kakumara: I think the `@ApiModelProperty` annotations here are overridden by the ones in the associated annotation ion the getter method. better to remove these and use only the one around getter.

kakumara:
Otherwise 👍

src/main/java/uk/gov/pay/api/model/PaymentState.java
ijmad: Sorted!

PR #96, opened 2016-05-13
PP-582 Add logging filter

imranbohoran:
## WHAT
Introducing a Logging filter that would log all incoming requests with a prefix. 
This gives us a nice trace of the request in the form of; start->error(if any)->end

## HOW TO REVIEW
Run through a payment journey and you should be able to see start->end logging for each resource hit on the connector.

## WHO
Any Dev

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/LoggingFilter.java
+++ b/src/main/java/uk/gov/pay/api/filter/LoggingFilter.java
@@ -1,0 +1,34 @@
+package uk.gov.pay.api.filter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+public class LoggingFilter implements Filter {
+
+    private static final Logger logger = LoggerFactory.getLogger(LoggingFilter.class);
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {}
+
+    @Override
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/api/filter/LoggingFilter.java
```diff
@@ -0,0 +1,34 @@
...
+        } catch (Throwable throwable){
+            logger.error("Exception - publicapi request - "+ requestURL + " - exception - "+ throwable.getMessage(), throwable);
+        }
+        finally {
+            logger.info("End - publicapi request - " + requestURL);
+        }
+    }
+
+    @Override
+    public void destroy() {}
```
kakumara: same comment as in connector. add a  log message here

src/main/java/uk/gov/pay/api/filter/LoggingFilter.java
```diff
@@ -0,0 +1,34 @@
+package uk.gov.pay.api.filter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+public class LoggingFilter implements Filter {
```
kakumara: This one and the associated  test can easily be in a common repository so that we don't need to copy/paste code acorss repos.

kakumara:
otherwise 👍

src/main/java/uk/gov/pay/api/filter/LoggingFilter.java
imranbohoran: 👍  Will add this as part of the next PR

PR #98, opened 2016-05-17
PP 335 Publicapi pagination and navigation hal links on payment search

rahul-vinayak:
- public api payments search now accepts page and display_size params, defaults to 1 and 500 respectively.
- Responses with the pagination info total, page, display_size, count and logical navigation links (prev/next/first/last) using the HAL terminology
- Navigation links are as received from the connector and modified accordingly


rauligs:
Would be good for all of us to follow new PR template style

src/main/java/uk/gov/pay/api/exception/mapper/SearchChargesExceptionMapper.java
```diff
@@ -9,19 +9,29 @@
...
 
 public class SearchChargesExceptionMapper implements ExceptionMapper<SearchChargesException> {
-
     private static final Logger LOGGER = LoggerFactory.getLogger(SearchChargesExceptionMapper.class);
 
     @Override
     public Response toResponse(SearchChargesException exception) {
-        PaymentError paymentError = aPaymentError(SEARCH_PAYMENTS_CONNECTOR_ERROR);
-        LOGGER.error("Connector invalid response was {}.\n Returning http status {} with error body {}", exception.getMessage(), INTERNAL_SERVER_ERROR, paymentError);
+        PaymentError paymentError = null;
```
rauligs: paymentError is not being use

src/main/java/uk/gov/pay/api/resources/HalResourceBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+            builder.addProperty(key, propertyMap.get(key));
+        }
+        return builder;
+    }
+
+    public String build() {
+        try {
+            return representationBuilder().build().serialize();
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
```
rauligs: Log the exception and pass the exception into it rather than doing a printStackTrace

src/main/java/uk/gov/pay/api/validation/PaymentSearchValidator.java
```diff
@@ -18,19 +18,33 @@
...
+        validatePage(pageNumber, validationErrors);
+        validateDisplaySize(displaySize, validationErrors);
 
         if (!validationErrors.isEmpty()) {
             throw new ValidationException(aPaymentError(SEARCH_PAYMENTS_VALIDATION_ERROR, join(validationErrors, ", ")));
         }
     }
 
+    private static void validatePage(String pageNumber, List<String> validationErrors) {
+        if (pageNumber!=null && Long.valueOf(pageNumber) < 1) {
```
rauligs: use isNotBlank() better?

src/test/java/uk/gov/pay/api/validation/PaymentSearchValidatorTest.java
```diff
@@ -14,46 +14,44 @@
...
+        PaymentSearchValidator.validateSearchParameters("confirmed", "ref", "2016-01-25T13-23:55Z", "2016-01-25T13:23:55Z", "1", "1");
     }
 
     @Test
     public void validateParams_shouldGiveAnErrorValidation_forAllParams() throws Exception {
+        expectedException.expect(aValidationExceptionContaining("P0401", "Invalid parameters: state, reference, from_date, to_date, page, display_size. See Public API documentation for the correct data formats"));
+        PaymentSearchValidator.validateSearchParameters("invalid", randomAlphanumeric(500), "2016-01-25T13-23:55Z", "2016-01-25T13-23:55Z", "-1", "-1");
+    }
 
+    @Test
```
rauligs: It seems some more tests could be added contemplating edge scenario cases with invalid inputs for new parameters

rauligs:
Most of the comments are minor changes, 👍  in general

src/main/java/uk/gov/pay/api/exception/mapper/SearchChargesExceptionMapper.java
rahul-vinayak: Fine to use, but I don't find it of much use, i.e. new template, people are just filling what and other info is of no use. Above is what explains what the PR is about

rahul-vinayak: paymentError is removed

src/main/java/uk/gov/pay/api/resources/HalResourceBuilder.java
rahul-vinayak: good point, done...

src/main/java/uk/gov/pay/api/validation/PaymentSearchValidator.java
rahul-vinayak: * It was a Long to start with thats why left with !=null, but isNotBlank makes sense with String.
* Integer now
* Will throw a NumberFormatException which is a Runtime exception if its not a number

src/test/java/uk/gov/pay/api/validation/PaymentSearchValidatorTest.java
rahul-vinayak: Can you be more specific on those scenarios

rauligs: For page and display size there are more failing scenarios for sure.
Is expected a long so I would add a test with a non long value (I would change it to Integer).
Is expected a long from 0? from 1? to n? n has a max value (max Integer value)
I would check for nonBlank rather than null.

rahul-vinayak: Is expected a long so I would add a test with a non long value (I would change it to Integer).
- _already changed to Integer, and have a test for non numeric value_ 
Is expected a long from 0? from 1? to n? n has a max value (max Integer value)
_added a test for max value +1 now_
I would check for nonBlank rather than null.
_already done that fix and there is a test for nulls_

PR #99, opened 2016-05-19
PP-541 Begin swap of Confirmed/Captured to Success

ijmad:
## WHAT
We allow the status 'success' to pass through publicapi without failing
validation. The Confirmed/Capture states will be removed after the changes
are applied to SelfService and Connector.

## HOW 
Add success to the list of allowed statuses. Switch some integration tests to 'success' so they will continue to be good tests after confirmed is removed.

## WHO
_Who should review this pull request:_

- [x] Developers
- [ ]  WebOps

### DIFF
--- a/src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
+++ b/src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
@@ -180,7 +180,7 @@ public class PaymentsResource {
                                    @Auth String accountId,
                                    @ApiParam(value = "Your payment reference to search", hidden = false)
                                    @QueryParam(REFERENCE_KEY) String reference,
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -180,7 +180,7 @@ public Response searchPayments(@ApiParam(value = "accountId", hidden = true)
                                    @Auth String accountId,
                                    @ApiParam(value = "Your payment reference to search", hidden = false)
                                    @QueryParam(REFERENCE_KEY) String reference,
-                                   @ApiParam(value = "State of payments to be searched. Example=confirmed", hidden = false, allowableValues = "range[created,started,submitted,failed,cancelled,error,confirmed,captured")
+                                   @ApiParam(value = "State of payments to be searched. Example=success", hidden = false, allowableValues = "range[created,started,submitted,success,failed,cancelled,error")
```

PR #103, opened 2016-05-24
PP-335 Added a check for non numeric page, display_size passed

rahul-vinayak:
- When non numeric value for page and display_size was passed it was
handling the error however validation was not populating the proper
error message, this fixes it, with a test


mrlumbu:
👍

PR #105, opened 2016-07-14
PP-800 Support for all supported HTTP verbs in request-denied resource

rauligs:
## WHAT
- When _Naxsi_ triggers a redirect (rule violated), the request to _/request-denied_ will use the same verb as the origin request so added the other verbs for this resource.

## WHO
- [X] Developers
- [X] WebOps


PauloPortugal:
:+1:

PR #107, opened 2016-07-29
[REVIEW ONLY] PP-926 alpine linux

ijmad:
## WHAT

Preliminary test of exchanging debian base with alpine for slimmer image with less attack surface for vulnerabilities.

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
-FROM java:8-jre
+FROM java:8-jre-alpine
+
+RUN apk add --no-cache bash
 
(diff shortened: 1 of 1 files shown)


ijmad:
rebuild

bazbremner:
I'd argue this isn't relevant to PP-926 - that's about directing alerts from security scans of containers, not reducing the attack service or selecting container base OSes.

Let's at least push this into another ticket, but I'm reluctant to sink too much time into this pre-MVP.

ijmad:
rebuild

PR #108, opened 2016-08-08
PP-942 Public API Refunds

rahul-vinayak:
- Resource for getting all refunds for a payment.
- Resource for getting a refund for a payment by ID.
- Resource for creating a new refund for a payment.
- Exception mapping for refunds.
- Refund validations for amount and validations returned from connector.
- Links added (as HAL resources), for search payments.
- Map error body from connector.

with @rauligs


src/main/java/uk/gov/pay/api/exception/mapper/CancelPaymentExceptionMapper.java
```diff
@@ -2,7 +2,7 @@
```
mrlumbu: being pedantic, should this be renamed CancelChargeExceptionMapper now?

src/test/java/uk/gov/pay/api/it/PaymentRefundsResourceITest.java
```diff
@@ -0,0 +1,183 @@
...
+                .body("_links.payment.href", is(paymentLocationFor(CHARGE_ID)))
+                .body("_embedded.refunds.size()", is(2))
+                .body("_embedded.refunds[0].refund_id", is("100"))
+                .body("_embedded.refunds[0].created_date", is(CREATED_DATE))
+                .body("_embedded.refunds[0].amount", is(100))
+                .body("_embedded.refunds[0].status", is("available"))
+                .body("_embedded.refunds[0]._links.size()", is(2))
+                .body("_embedded.refunds[0]._links.self.href", is(paymentRefundLocationFor(CHARGE_ID, "100")))
+                .body("_embedded.refunds[0]._links.payment.href", is(paymentLocationFor(CHARGE_ID)));
+    }
```
mrlumbu: Would prefer if you checked the second refund too

mrlumbu:
two minor comments, otherwise LGTM 👍

src/main/java/uk/gov/pay/api/exception/mapper/CancelPaymentExceptionMapper.java
rahul-vinayak: Probably yes...

src/test/java/uk/gov/pay/api/it/PaymentRefundsResourceITest.java
rahul-vinayak: idea of the test was it would break if the structure of refund changes, but can add the extra checks if it helps

mrlumbu:
👍

PR #112, opened 2016-08-11
PP-864 search payments by email

rahul-vinayak:
- Public api now returns email in the response for search payments, get
payment by ID
- Payments search can be filtered on email address.
with @PauloPortugal


ckalista:
Reviewed.

PR #113, opened 2016-08-16
PP-864 We should not serialize email field if null

PauloPortugal:
## WHAT
 * We should not serialize email field from the payment response if it is null

## HOW 
_Steps to test or reproduce:_

```
cd $WORKSPACE
msl reset && msl -a up && ./run-endtoend.sh
```


PauloPortugal:
@rauligs  Please verity you are happy with these changes. Thank you :)

rauligs:
👍

PR #114, opened 2016-09-01
PP-720 Pass through conflict state from connector

mrlumbu:
Connector has been updated to return 409 CONFLICT when charge cancellation attempted on charge  in lock state. This commit ensures the correct code is passed on to the world.


rauligs:
👍

PR #115, opened 2016-09-19
PP-1155 Removed email when logging payments

rauligs:
## WHAT
- Personal information shouldn't be logged. Removed email from Payment.toString() to avoid logging unwanted data to be displayed.


rauligs:
rebuild

PR #116, opened 2016-09-19
Pp 863 card types used available to partners

SandorArpa:
The used card brand, eg visa, master-card, etc is available for search as card_brand. The API returns the Label of the card brand, eg Visa, Mastercard, etc with charges status greater than 'CAPTURE SUBMITTED'

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/ChargeFromResponse.java
+++ b/src/main/java/uk/gov/pay/api/model/ChargeFromResponse.java
@@ -21,7 +21,7 @@ public class ChargeFromResponse {
     private String paymentProvider;
 
     @JsonProperty("card_brand")
-    private String cardLabel;
+    private String cardBrand;
 
     @JsonProperty("links")
     private List<PaymentConnectorResponseLink> links = new ArrayList<>();
@@ -68,7 +68,7 @@ public class ChargeFromResponse {
         return paymentProvider;
     }
 
-    public String getCardLabel() { return cardLabel; }
+    public String getCardBrand() { return cardBrand; }
 
     public String getCreated_date() {
         return created_date;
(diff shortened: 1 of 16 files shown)


ckalista:
rebuild

ckalista:
rebuild

ckalista:
build

ckalista:
rebuild

src/main/java/uk/gov/pay/api/model/ChargeFromResponse.java
```diff
@@ -20,6 +20,9 @@
     @JsonProperty("payment_provider")
     private String paymentProvider;
 
+    @JsonProperty("card_brand")
+    private String cardLabel;
```
rahul-vinayak: if (brand != label) 
        "pick a better property"

src/main/java/uk/gov/pay/api/model/PaymentForSearchResult.java
```diff
@@ -40,6 +40,7 @@ public static PaymentForSearchResult valueOf(
                 paymentResult.getReference(),
                 paymentResult.getEmail(),
                 paymentResult.getPaymentProvider(),
+                paymentResult.getCardLabel(),
```
rahul-vinayak: are label and brand used for same thing?

rahul-vinayak:
since the email and reference searches have been made case-insensitive while searching, would be good to make sure the card type is case-insensitive as well, make sure connector end point would accept both "Visa" as well as "visa". So would be be good to write a test for that, and the required change is in the connector layer makes it case-insensitive, like the ref and email.

src/main/java/uk/gov/pay/api/model/ChargeFromResponse.java
SandorArpa: I have addressed the issue. Renamed variable accordingly.

SandorArpa:
rebuild

src/main/java/uk/gov/pay/api/model/PaymentForSearchResult.java
SandorArpa: I have addressed the issue. Renamed methods accordingly.

SandorArpa:
rebuild

rahul-vinayak:
👍

PR #120, opened 2016-09-21
Swagger hostname update

davidillsley:
## WHAT
Update the swagger config and generated JSON file to point to the production environment.


bazbremner:
I'm not going to wait on tests for that change. Merged, thanks.

PR #121, opened 2016-09-27
PP-1219 Fix logging of payment errors

rauligs:
## WHAT
- Added name and code value to toString() method so we can get same the information our integrators receive (code value)


ijmad:
:+1:

PR #125, opened 2016-11-17
PP-1282 removing county field and updating api documentation

kakumara:
## WHAT
- County is not a field that we capture/persist. so removing it.
- Updating gelato documentation.


kakumara:
rebuild

PauloPortugal:
:+1:

PR #126, opened 2016-11-30
PP-1386 Add dropwizard metrics

imranbohoran:
## WHAT

- We added two environment variables for host and port for
metrics.
- If the variables are not passed, default values are going to
get picked up

with @imranbohoran


ckalista:
Code review completed.

PR #127, opened 2016-12-02
PP-1452 Update documentation

rauligs:
## WHAT
- Fix documentation, where message is "Accepted" when creating refunds the correct response code is 202 (not 201).


simad:
rebuild

kakumara:
LGTM 👍

PR #130, opened 2017-01-20
PP-1517 Pass settlement_summary through PublicApi

simad:
## WHAT
- `settlement_summary` is a structure coming from connector, containing
  two fields: `capture_submit_time`, which is the timestamp of us
  receiving the payment gateway's notification of a capture, and
  `captured_date`, which is the date contained in the notification
  payload.
- if a charge has not been captured, both those dates  could potentially
  be empty; connector would be returning null.
  In this case, public api would not show at all the keys in the json structure
- `settlement_summary` is serialised in calls to `/payments` and
  `/payments/{id}` endpoints

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,10 @@ Content-Type: application/json
         "amount_available": 14500
         "amount_submitted": 0
     }
+    "settlement_summary": {
+        "captured_date": "2016-01-15",
+        "capture_submit_time": "2016-01-15T16:30:56Z" 
+    }
 }
 ```
 
@@ -169,6 +173,8 @@ Content-Type: application/json
 | `refund_summary.status`| Refund availability status of the payment                                                |
 | `refund_summary.amount_available`| Amount available for refunds                                                   |
 | `refund_summary.amount_submitted`| Total amount of refunds submitted for this payment                             |
+| `settlement_summary.captured_date`| Date of the capture according to the payment gateway                          |
+| `settlement_summary.capture_submit_time`| Date and time of submission of the capture request, if present          |
 | `created_date`         | The payment creation date for this payment                                               |
 | `_links.self`          | Link to the payment                                                                      |
 | `_links.next_url`      | Where to navigate the user next as a GET                                                 |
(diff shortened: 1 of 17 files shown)


src/test/java/uk/gov/pay/api/it/fixtures/PaymentSearchResultBuilder.java
```diff
@@ -21,6 +21,8 @@
     private static final int DEFAULT_NUMBER_OF_RESULTS = 3;
 
     public static final String DEFAULT_CREATED_DATE = DateTimeUtils.toUTCDateString(ZonedDateTime.now());
+    public static final String DEFAULT_CAPTURE_SUBMIT_TIME = DateTimeUtils.toUTCDateString(ZonedDateTime.now());
```
ckalista: I would imagine it is deliberate to convert the zoned date time differently?

README.md
```diff
@@ -150,6 +150,10 @@ Content-Type: application/json
         "amount_available": 14500
         "amount_submitted": 0
     }
+    "settlement_summary": {
+        "captured_date": "2016-01-15",
```
ckalista: Does the format of both dates need to be different?

src/test/java/uk/gov/pay/api/it/fixtures/PaymentSearchResultBuilder.java
simad: Yes, one of the requirements is that everything should be UTC - and that's what's coming from connector, so the conversion should not actually change the time zone.

README.md
simad: Unfortunately, yes - Worldpay does not return the time of capture, just the date. It's returned to us as `bookingDate` in the payload of the "payment captured" notification. It's true that we could've expressed this date as `2016-01-15T00:00:00Z`, for example, but I guess rather than returning potentially misleading information, it's better to stick to what we know for sure.

PR #131, opened 2017-01-25
Upgrade dependencies

ijmad:
Upgrade jackson and commons-validator versions
Remove old/unused/broken newrelic JARs from source tree


kakumara:
👍  LGTM

PR #141, opened 2017-05-09
PP-1940 Don’t log payment descriptions because they may contain PII

alexbishop1:
Some services put personally-identifiable information into payment descriptions. In order to avoid this ending up in the logs, exclude the description when logging payment details.


bazbremner:
Looks fine to me. Avoiding merging until ZAP tests have been resolved.

PR #142, opened 2017-05-17
PP-2057 document that capture_submit_time can be null

heathd:
If we don't get an acknowledgement from a payment gateway of a capture order,
but that gateway does successfully process the order, then a payment can be
`CAPTURED` but `capture_submit_time` will be null.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/SettlementSummary.java
+++ b/src/main/java/uk/gov/pay/api/model/SettlementSummary.java
@@ -24,7 +24,7 @@ public class SettlementSummary {
         this.capturedDate = capturedDate;
     }
 
-    @ApiModelProperty(value = "Date and time capture request has been submitted", example = "2016-01-21T17:15:00Z")
+    @ApiModelProperty(value = "Date and time capture request has been submitted (may be null if capture request was not immediately acknowledged by payment gateway)", example = "2016-01-21T17:15:00Z")
     public String getCaptureSubmitTime() {
         return captureSubmitTime;
     }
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/api/model/SettlementSummary.java
```diff
@@ -24,7 +24,7 @@ public SettlementSummary(String captureSubmitTime, String capturedDate) {
         this.capturedDate = capturedDate;
     }
 
-    @ApiModelProperty(value = "Date and time capture request has been submitted", example = "2016-01-21T17:15:00Z")
+    @ApiModelProperty(value = "Date and time capture request has been submitted (may be null if capture request was not acknowledged by payment gateway)", example = "2016-01-21T17:15:00Z")
```
alexbishop1: This covers the case where the capture request times out but they then send us a notification saying it was captured, right? Perhaps “may be null if capture request was not immediately acknowledged by payment gateway” (note “immediately”) would be a little clearer.

swagger/swagger.json
```diff
@@ -767,7 +767,7 @@
         "capture_submit_time" : {
           "type" : "string",
           "example" : "2016-01-21T17:15:00Z",
-          "description" : "Date and time capture request has been submitted",
+          "description" : "Date and time capture request has been submitted (may be null if capture request was not acknowledged by payment gateway)",
```
alexbishop1: Don’t forget the Swagger!

heathd: oops, thanks.

Updated and squashed the commits

PR #147, opened 2017-08-14
PP-2514 Enable https validation for return url

DanailMinchev:
## WHAT

We need to return client error HTTP status code in case we submit `return_url` which is non-HTTPS (currently `NAXSI` is returning `404` in case we submit non-HTTPS url).
We use `422 Unprocessable Entity` for client errors.

## HOW 

- Added `allowHttpForReturnUrl` setting (default is `false`) which enables or disables "https only" check for `return_url` request parameter
- Added related tests in `PaymentRequestValidatorTest` and modified existing ones
- Updated documentation

### DIFF
--- a/src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java
+++ b/src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java
@@ -13,34 +13,80 @@ public class PaymentRequestValidatorTest {
 
     @Rule
     public ExpectedException expectedException = ExpectedException.none();
-    private static final String SUCCESSFUL_TEST_EMAIL = "alice.111@mail.fake";
-    private static final String UNSUCCESSFUL_TEST_EMAIL = randomAlphanumeric(255) + "@mail.fake";
+
+    private static final int VALID_AMOUNT = 100;
+    private static final String VALID_RETURN_URL = "https://www.example.com/return_url";
+    private static final String VALID_REFERENCE = "reference";
+    private static final String VALID_DESCRIPTION = "description";
+
+    private final PaymentRequestValidator paymentRequestValidator = new PaymentRequestValidator(URLValidator.SECURITY_ENABLED);
 
     @Test
-    public void validateMinimumAmount_shouldSuccessValue(){
-        PaymentRequestValidator paymentRequestValidator = new PaymentRequestValidator(URLValidator.SECURITY_ENABLED);
-        CreatePaymentRequest paymentRequest = new CreatePaymentRequest(1, "https://return.to/return_url", "reference", "description");
+    public void validParameters_shouldSuccessValue() {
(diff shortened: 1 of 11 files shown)


src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java
```diff
@@ -13,34 +13,80 @@
...
+    public void validateMinimumAmount_shouldSuccessValue() {
+        CreatePaymentRequest createPaymentRequest = createPaymentRequest(PaymentRequestValidator.AMOUNT_MIN_VALUE, VALID_RETURN_URL, VALID_REFERENCE, VALID_DESCRIPTION);
+        paymentRequestValidator.validate(createPaymentRequest);
+    }
 
     @Test
-    public void validateMinimumAmount_shouldSuccessValue(){
-        PaymentRequestValidator paymentRequestValidator = new PaymentRequestValidator(URLValidator.SECURITY_ENABLED);
-        CreatePaymentRequest paymentRequest = new CreatePaymentRequest(1, "https://return.to/return_url", "reference", "description");
+    public void validateMinimumAmount_shouldFailValue() throws Exception {
```
alexbishop1: I assume what this test tests for has been intentionally changed because it wrong previously. (Before it claimed to be checking going below the minimum amount causes a validation failure but it was actually checking going above the maximum amount causes a validation failure.)

src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java
```diff
@@ -13,34 +13,80 @@
...
+        CreatePaymentRequest paymentRequest = createPaymentRequest(VALID_AMOUNT, invalidMaxLengthReturnUrl, VALID_REFERENCE, VALID_DESCRIPTION);
+        expectedException.expect(ValidationException.class);
+        paymentRequestValidator.validate(paymentRequest);
     }
 
     @Test
-    public void validateMaximumAmount_shouldFailValue() throws Exception{
-        PaymentRequestValidator paymentRequestValidator = new PaymentRequestValidator(URLValidator.SECURITY_ENABLED);
-        CreatePaymentRequest paymentRequest = new CreatePaymentRequest(PaymentRequestValidator.AMOUNT_MAX_VALUE + 1, "https://return.to/return_url", "reference", "description");
+    public void validateReturnUrlSecurityEnabled_shouldFailValue() throws Exception {
```
alexbishop1: Can we have a more precise name like `validateReturnUrlNotHttps_shouldFailValue()` or something?

DanailMinchev: 👍

src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java
DanailMinchev: Yes, that is correct

PR #154, opened 2017-11-13
PP-2888 Add direct debit payments support

simad:
## WHAT
 - We are introducing a new payment type, Direct Debit, as opposed to Card payments. The create a payment api is transparent to this change, but the internal representation (and the entity returned to the user) is different.
- The `next_url` of a payment is built using a specific connector, dd/cc depending on the payment type. This is inferred by public api by a `token_type` field added to public auth (https://github.com/alphagov/pay-publicauth/pull/66)
 - A few credit card-specific fields have been moved to `CardPayment`. An example is card brand information or settlement summary
 - The current iteration does not support cancel / refund / events links for Direct Debit payments.
 - The inheritance model for `PaymentWithAllLinks` has been changed, it nows uses composition instead of pure inheritance. It _has_ a payment and some links, instead of _extending_ a payment to add links. To maintain backwards compatibility, the `payment` object inside `PaymentWithAllLinks` is unwrapped, its properties are assigned to the top level object
 - Note: TokenPaymentType is a duplicate of the same class in PublicAuth and should be moved to the shared  module repository once we have one


simad:
rebuild

simad:
rebuild

src/main/resources/config/config.yaml
```diff
@@ -17,6 +17,7 @@ logging:
 
 baseUrl: ${PUBLICAPI_BASE}
 connectorUrl: ${CONNECTOR_URL}
```
DanailMinchev: Maybe we should rename:
- `connectorUrl -> connectorCCUrl`
- `CONNECTOR_URL -> CONNECTOR_CC_URL`
as part of clean up or refactoring in the future

DanailMinchev:
LGTM 👍 
Needs rebasing

PauloPortugal:
Awesome work @simad!!!

PR #177, opened 2018-05-22
Pp 3733 pact consumer test

oswaldquek:
## WHAT
A working pact consumer java integration test. It was agreed with @alexbishop1 and @mrlumbu we should make the direct debit payment test use pact. See commit messages for more detail.


oswaldquek:
DO NOT MERGE YET

PR #178, opened 2018-05-24
PP-3841: Publish pacts

oswaldquek:
## WHAT
Add the pact maven plugin to enable publishing of pacts.
Set trimSnapshot to true in the pom.xml as the pact broker has issues parsing
SNAPSHOT versions (even though we aren't using SNAPSHOT versions but rather the
git commit sha, but just in case). See
https://github.com/DiUS/pact-jvm/tree/master/pact-jvm-provider-maven#publishing-pact-files-to-a-pact-broker.

For some reason the pact broker homepage (https://pact-broker-test.cloudapps.digital/) doesn't show the latest publicapi pact that was published. But the latest version that was published from the https://build.ci.pymnt.uk/job/pay-publicapi/job/PR-178/3 is definitely there at https://pact-broker-test.cloudapps.digital/hal-browser/browser.html#/pacts/provider/direct-debit-connector/consumer/publicapi/version/1fb63718d85705ab17d2e032e84ad49c1d5c5fbf.

@oswaldquek

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,8 +28,15 @@ pipeline {
       steps {
         script {
           def long stepBuildTime = System.currentTimeMillis()
+          def commit = gitCommit()
 
-          sh 'mvn clean package'
+          withCredentials([
+                  string(credentialsId: 'pact_broker_username', variable: 'PACT_BROKER_USERNAME'),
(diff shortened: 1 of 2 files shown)


oswaldquek:
DO NOT MERGE

PR #181, opened 2018-05-25
PP-3676: Added Agreement skeleton resource and entities

georgeracu:
PP-3676 Public API part of create `Agreement`

- Added resource POST `/v1/agreement`
- Added objects for domain model entity and JSON request and response
- Mocked Direct Debit Connector mock that should accept a POST to
`/v1/accounts/{account_id}/agreements`
- Added Pact test instead of e2e


src/main/java/uk/gov/pay/api/exception/mapper/CreateAgreementExceptionMapper.java
```diff
@@ -0,0 +1,39 @@
...
+public class CreateAgreementExceptionMapper implements ExceptionMapper<CreateAgreementException> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(CreateAgreementExceptionMapper.class);
+
+    @Override
+    public Response toResponse(CreateAgreementException exception) {
+
+        PaymentError paymentError;
+
+        if (exception.getErrorStatus() == NOT_FOUND.getStatusCode()) {
```
simad: this might look a bit better with a ternary operator ie.
``` 
PaymentError paymentError = exception.getErrorStatus() == NOT_FOUND.getStatusCode() ? 
aPaymentError(CREATE_PAYMENT_ACCOUNT_ERROR) : aPaymentError(CREATE_PAYMENT_CONNECTOR_ERROR)
```

PR #184, opened 2018-06-12
PP-3676 Fixed attribute `value` Swagger error

DanailMinchev:
## WHAT

- When we have space in the name we receive `$ref values must be RFC3986-compliant percent-encoded URIs` error

with @georgeracu


DanailMinchev:
@alexbishop1 yes, we could add `%20`, but we actually decided to go with default behaviour (the name of the class) as described in `io.swagger.annotations.ApiModel#value`  : 
```
    /**
     * Provide an alternative name for the model.
     * <p>
     * By default, the class name is used.
     */
```

PR #185, opened 2018-06-13
Guice publicapi

alexbishop1:
Start using Guice to instantiate objects added to the environment in the `PublicApi` class.

This will make it easier to extract some of the logic in the `*Resource` classes to separate service classes (because we will be able to use Guice to inject their dependencies).

with @heathd

### DIFF
--- a/src/main/java/uk/gov/pay/api/app/config/PublicApiModule.java
+++ b/src/main/java/uk/gov/pay/api/app/config/PublicApiModule.java
@@ -1,0 +1,22 @@
+package uk.gov.pay.api.app.config;
+
+import com.google.inject.AbstractModule;
+import io.dropwizard.setup.Environment;
+
+public class PublicApiModule extends AbstractModule {
+
+    private final PublicApiConfig configuration;
+    private final Environment environment;
+
+    public PublicApiModule(final PublicApiConfig configuration, final Environment environment) {
+        this.configuration = configuration;
+        this.environment = environment;
+    }
+
+    @Override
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/api/resources/PaymentRefundsResource.java
```diff
@@ -55,10 +72,11 @@
     private final Client client;
     private final String connectorUrl;
 
-    public PaymentRefundsResource(String baseUrl, Client client, String connectorUrl) {
-        this.baseUrl = baseUrl;
+    @Inject
+    public PaymentRefundsResource(Client client, PublicApiConfig configuration) {
```
simad: 👍 for passing the config around

src/main/java/uk/gov/pay/api/app/config/PublicApiModule.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public PublicApiModule(final PublicApiConfig configuration, final Environment environment) {
+        this.configuration = configuration;
+        this.environment = environment;
+    }
+
+    @Override
+    protected void configure() {
+        bind(PublicApiConfig.class).toInstance(configuration);
+        bind(Environment.class).toInstance(environment);
```
simad: I assume you considered if there's anything we want to be bound as `Singleton`, so 👍

simad: Nicely spotted

PR #186, opened 2018-06-14
Extract create payments service

alexbishop1:
with @heathd 

Extract the body of the 'create payment' api call into a service object. We
followed these principles:

1. 'presentation' concerns such as serialisation of the response and generation
   of headers remain in the Resource
2. issuing the call to connector and parsing the response is done in the service object

We also extracted `PublicApiUriGenerator` and `ConnectorUriGenerator` which are
responsible for generating URIs relating to PublicApi and Connector
respectively. They get the dropwizard app configuration injected to their
constructors.

Finally we created a Pact contract test for the `CreatePaymentService`. Where
possible we've used real objects where they are simple value objects, rather
than using mocks.

## Inline all resource urls and field name constants 

The url paths are almost exclusively used only in one place, which is the
configuration of the resource endpoint. It's much clearer to have the string
literal visible on the resource annotation than to have it obscured by
constant.

We also inlined some constants which define the keys from request payloads.
These are also part of our external API, not likely to change and used only in
one or two places. Again, having a string literal makes the code easier to read
in these cases.

## rename Account.getName() to Account.getAccountId()

the field actually contains the account id, so it's confusing to call it name.

the `getName()` method is used when the Account object represents a principal in the authentication system, so add a `getName()` method which overrides that and delegates to `getAccountId()`.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/PaymentState.java
+++ b/src/main/java/uk/gov/pay/api/model/PaymentState.java
@@ -91,7 +91,6 @@ public class PaymentState {
 
     @Override
     public int hashCode() {
-
         return Objects.hash(status, finished, message, code);
     }
 }
--- a/src/main/java/uk/gov/pay/api/model/PaymentState.java
+++ b/src/main/java/uk/gov/pay/api/model/PaymentState.java
@@ -7,6 +7,8 @@ import com.fasterxml.jackson.databind.JsonNode;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 
+import java.util.Objects;
+
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
 @ApiModel(value="PaymentState", description = "A structure representing the current state of the payment in its lifecycle.")
@@ -33,6 +35,10 @@ public class PaymentState {
     }
 
     public PaymentState() {
+    }
+
+    public PaymentState(String status, boolean finished) {
+        this(status, finished, null, null);
     }
 
     public PaymentState(String status, boolean finished, String message, String code) {
@@ -70,5 +76,22 @@ public class PaymentState {
                 ", message=" + message +
                 ", code=" + code +
                 '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        PaymentState that = (PaymentState) o;
+        return finished == that.finished &&
+                Objects.equals(status, that.status) &&
+                Objects.equals(message, that.message) &&
+                Objects.equals(code, that.code);
+    }
+
+    @Override
+    public int hashCode() {
+
+        return Objects.hash(status, finished, message, code);
     }
 }
--- a/src/main/java/uk/gov/pay/api/model/links/Link.java
+++ b/src/main/java/uk/gov/pay/api/model/links/Link.java
@@ -58,7 +58,6 @@ public class Link {
 
     @Override
     public int hashCode() {
-
         return Objects.hash(href, method);
     }
 }
--- a/src/main/java/uk/gov/pay/api/model/links/Link.java
+++ b/src/main/java/uk/gov/pay/api/model/links/Link.java
@@ -5,6 +5,8 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 
+import java.util.Objects;
+
 import static com.fasterxml.jackson.annotation.JsonInclude.Include;
 
 @ApiModel(value = "Link", description = "A link related to a payment")
@@ -16,7 +18,7 @@ public class Link {
     @JsonProperty(value = "method")
     private String method;
 
-    Link(String href, String method) {
+    public Link(String href, String method) {
         this.href = href;
         this.method = method;
     }
@@ -43,5 +45,20 @@ public class Link {
                 "href='" + href + '\'' +
                 ", method='" + method + '\'' +
                 '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Link link = (Link) o;
+        return Objects.equals(href, link.href) &&
+                Objects.equals(method, link.method);
+    }
+
+    @Override
+    public int hashCode() {
+
+        return Objects.hash(href, method);
     }
 }
--- a/src/main/java/uk/gov/pay/api/model/links/PostLink.java
+++ b/src/main/java/uk/gov/pay/api/model/links/PostLink.java
@@ -63,7 +63,6 @@ public class PostLink extends Link {
 
     @Override
     public int hashCode() {
-
         return Objects.hash(super.hashCode(), type, params);
     }
 }
--- a/src/main/java/uk/gov/pay/api/model/links/PostLink.java
+++ b/src/main/java/uk/gov/pay/api/model/links/PostLink.java
@@ -5,6 +5,7 @@ import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 
 import java.util.Map;
+import java.util.Objects;
 
 import static com.fasterxml.jackson.annotation.JsonInclude.Include;
 
@@ -15,13 +16,13 @@ public class PostLink extends Link {
     private String type;
     private Map<String, Object> params;
 
-    PostLink(String href, String method, String type, Map<String, Object> params) {
+    public PostLink(String href, String method, String type, Map<String, Object> params) {
         super(href, method);
         this.type = type;
         this.params = params;
     }
 
-    PostLink(String href, String method) {
+    public PostLink(String href, String method) {
         super(href, method);
     }
 
@@ -48,5 +49,21 @@ public class PostLink extends Link {
                 ", type='" + type + '\'' +
                 ", params=" + params +
                 '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        if (!super.equals(o)) return false;
+        PostLink postLink = (PostLink) o;
+        return Objects.equals(type, postLink.type) &&
+                Objects.equals(params, postLink.params);
+    }
+
+    @Override
+    public int hashCode() {
+
+        return Objects.hash(super.hashCode(), type, params);
     }
 }
--- a/src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
+++ b/src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
@@ -77,9 +77,6 @@ public class PaymentsResource {
     private final PublicApiUriGenerator publicApiUriGenerator;
     private final ConnectorUriGenerator connectorUriGenerator;
 
-    private final String connectorUrl;
-    private final String connectorDDUrl;
-
     @Inject
     public PaymentsResource(Client client, CreatePaymentService createPaymentService,
                             ObjectMapper objectMapper, PublicApiConfig configuration,
@@ -87,8 +84,6 @@ public class PaymentsResource {
         this.client = client;
         this.createPaymentService = createPaymentService;
         this.baseUrl = configuration.getBaseUrl();
-        this.connectorUrl = configuration.getConnectorUrl();
-        this.connectorDDUrl = configuration.getConnectorDDUrl();
         this.objectMapper = objectMapper;
         this.publicApiUriGenerator = publicApiUriGenerator;
         this.connectorUriGenerator = connectorUriGenerator;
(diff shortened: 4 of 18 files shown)


src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -139,27 +109,35 @@ public PaymentsResource(Client client, ObjectMapper objectMapper, PublicApiConfi
...
-                               @PathParam(PAYMENT_KEY) String paymentId) {
+                               @PathParam("paymentId") String paymentId) {
 
         logger.info("Payment request - paymentId={}", paymentId);
+        String url = account.getPaymentType().equals(DIRECT_DEBIT) ? connectorDDUrl : connectorUrl;
+        UriBuilder builder = UriBuilder
+                .fromPath(url)
+                .path(format("/v1/api/accounts/%s/charges/%s", account.getAccountId(), paymentId));
+
+        Collections.<Pair<String, String>>emptyList().stream().forEach(pair -> {
```
oswaldquek: are we streaming an empty list here?

heathd: will have a look...

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -139,27 +109,35 @@ public PaymentsResource(Client client, ObjectMapper objectMapper, PublicApiConfi
...
             @ApiResponse(code = 500, message = "Downstream system error", response = PaymentError.class)})
     public Response getPayment(@ApiParam(value = "accountId", hidden = true) @Auth Account account,
-                               @PathParam(PAYMENT_KEY) String paymentId) {
+                               @PathParam("paymentId") String paymentId) {
 
         logger.info("Payment request - paymentId={}", paymentId);
+        String url = account.getPaymentType().equals(DIRECT_DEBIT) ? connectorDDUrl : connectorUrl;
+        UriBuilder builder = UriBuilder
+                .fromPath(url)
+                .path(format("/v1/api/accounts/%s/charges/%s", account.getAccountId(), paymentId));
```
oswaldquek: this might be for a later PR, but i was wondering if this could be covered by a connector client or connector url generator too (similar to publicApiUriGenerator).

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -183,23 +161,31 @@ public Response getPayment(@ApiParam(value = "accountId", hidden = true) @Auth A
...
+                                     @PathParam("paymentId") String paymentId) {
 
         logger.info("Payment events request - payment_id={}", paymentId);
 
+        String url = account.getPaymentType().equals(DIRECT_DEBIT) ? connectorDDUrl : connectorUrl;
+        UriBuilder builder = UriBuilder
+                .fromPath(url)
+                .path(format("/v1/api/accounts/%s/charges/%s/events", account.getAccountId(), paymentId));
+
+        Collections.<Pair<String, String>>emptyList().stream().forEach(pair -> {
```
oswaldquek: streaming an empty list?

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -287,16 +280,17 @@ public Response searchPayments(@ApiParam(value = "accountId", hidden = true)
                 JsonNode responseJson = connectorResponse.readEntity(JsonNode.class);
                 logger.debug("json response from connector from charge search: " + responseJson);
 
-                TypeReference<PaymentSearchResponse> typeRef = new TypeReference<PaymentSearchResponse>() {};
+                TypeReference<PaymentSearchResponse> typeRef = new TypeReference<PaymentSearchResponse>() {
+                };
```
oswaldquek: this doesn't need to go on a newline

src/main/java/uk/gov/pay/api/service/CreatePaymentService.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Inject
+    public CreatePaymentService(Client client, PublicApiConfig configuration, PublicApiUriGenerator publicApiUriGenerator, ConnectorUriGenerator connectorUriGenerator) {
+        this.client = client;
+        this.baseUrl = configuration.getBaseUrl();
+        this.publicApiUriGenerator = publicApiUriGenerator;
+        this.connectorUriGenerator = connectorUriGenerator;
+    }
+
+    public PaymentWithAllLinks invoke(Account account, CreatePaymentRequest requestPayload) {
```
oswaldquek: i think `create` would be a better method name

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -404,14 +382,22 @@ public Response createNewPayment(@ApiParam(value = "accountId", hidden = true) @
...
+                                  @PathParam("paymentId") String paymentId) {
 
         logger.info("Payment cancel request - payment_id=[{}]", paymentId);
 
+        String url = account.getPaymentType().equals(DIRECT_DEBIT) ? connectorDDUrl : connectorUrl;
+        UriBuilder builder = UriBuilder
+                .fromPath(url)
+                .path(format("/v1/api/accounts/%s/charges/%s/cancel", account.getAccountId(), paymentId));
+
+        Collections.<Pair<String, String>>emptyList().stream().forEach(pair -> {
```
oswaldquek: streaming an empty list

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -423,37 +409,23 @@ public Response cancelPayment(@ApiParam(value = "accountId", hidden = true) @Aut
...
-    }
-
     private String getConnectorUrl(TokenPaymentType paymentType, String urlPath) {
-        return getConnectorUrl(paymentType, urlPath, Collections.emptyList());
+        String url = paymentType.equals(DIRECT_DEBIT) ? connectorDDUrl : connectorUrl;
+        UriBuilder builder = UriBuilder
+                .fromPath(url)
+                .path(urlPath);
+
+        Collections.<Pair<String, String>>emptyList().stream().forEach(pair -> {
```
oswaldquek: streaming an empty list

src/main/java/uk/gov/pay/api/service/ConnectorUriGenerator.java
```diff
@@ -0,0 +1,39 @@
...
+    private String connectorBaseUrlForAccount(Account account) {
+        return isDirectDebitAccount(account) ? configuration.getConnectorDDUrl() : configuration.getConnectorUrl();
+    }
+
+    private boolean isDirectDebitAccount(Account account) {
+        return account.getPaymentType().equals(DIRECT_DEBIT);
+    }
+
+    public URI getPaymentEventsURI(String baseUrl, String charge_id) {
+        return null;
```
oswaldquek: is this method needed if it's returning null?

src/main/java/uk/gov/pay/api/service/PublicApiUriGenerator.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.pay.api.service;
+
+import javax.ws.rs.core.UriBuilder;
+import java.net.URI;
+
+public class PublicApiUriGenerator {
```
oswaldquek: how about initialising PublicApiUriGenerator with a baseUrl? looks like baseUrl is always the same.

src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
```diff
@@ -0,0 +1,61 @@
...
+        createPaymentService = new CreatePaymentService(client, configuration, publicApiUriGenerator, connectorUriGenerator);
+    }
+
+    @Test
+    @PactVerification({"connector"})
+    @Pacts(pacts = {"publicapi-connector"})
+    public void testCreatePayment() {
+        Account account = new Account("GATEWAY_ACCOUNT_ID", TokenPaymentType.CARD);
+        CreatePaymentRequest requestPayload = new CreatePaymentRequest(100, "https://somewhere.gov.uk/rainbow/1", "a reference", "a description");
+        createPaymentService.invoke(account, requestPayload);
```
oswaldquek: be good to have some assertions...?

heathd: pact covers the external side-effect, but you're right, we should have an assertion about the return value of `invoke`. We got carried away with adding a pact :-)

src/main/java/uk/gov/pay/api/service/CreatePaymentService.java
alexbishop1: Renamed from `invoke` to `create`

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
alexbishop1: Newline squished

src/main/java/uk/gov/pay/api/service/ConnectorUriGenerator.java
alexbishop1: Now gone

src/main/java/uk/gov/pay/api/service/PublicApiUriGenerator.java
alexbishop1: We now get Guice to pass the `PublicApiConfig` to the `PublicApiUriGenerator` constructor, which retrieves the `baseUrl` from it and stores it in a field

src/main/java/uk/gov/pay/api/model/links/Link.java
```diff
@@ -44,4 +46,19 @@ public String toString() {
                 ", method='" + method + '\'' +
                 '}';
     }
+
+    @Override
+    public boolean equals(Object o) {
```
oswaldquek: might it not be easier to use https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/builder/EqualsBuilder.html?

src/main/java/uk/gov/pay/api/model/PaymentState.java
```diff
@@ -71,4 +77,21 @@ public String toString() {
                 ", code=" + code +
                 '}';
     }
+
+    @Override
+    public boolean equals(Object o) {
```
oswaldquek: might it not be easier to use https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/builder/EqualsBuilder.html?

src/main/java/uk/gov/pay/api/model/PaymentState.java
```diff
@@ -71,4 +77,21 @@ public String toString() {
...
+        PaymentState that = (PaymentState) o;
+        return finished == that.finished &&
+                Objects.equals(status, that.status) &&
+                Objects.equals(message, that.message) &&
+                Objects.equals(code, that.code);
+    }
+
+    @Override
+    public int hashCode() {
+
```
oswaldquek: empty line. sorry!

src/main/java/uk/gov/pay/api/model/links/Link.java
```diff
@@ -44,4 +46,19 @@ public String toString() {
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Link link = (Link) o;
+        return Objects.equals(href, link.href) &&
+                Objects.equals(method, link.method);
+    }
+
+    @Override
+    public int hashCode() {
+
```
oswaldquek: empty line

src/main/java/uk/gov/pay/api/model/links/PostLink.java
```diff
@@ -49,4 +50,20 @@ public String toString() {
                 ", params=" + params +
                 '}';
     }
+
+    @Override
+    public boolean equals(Object o) {
```
oswaldquek: see other comment on EqualsBuilder

src/main/java/uk/gov/pay/api/model/links/PostLink.java
```diff
@@ -49,4 +50,20 @@ public String toString() {
...
+        if (o == null || getClass() != o.getClass()) return false;
+        if (!super.equals(o)) return false;
+        PostLink postLink = (PostLink) o;
+        return Objects.equals(type, postLink.type) &&
+                Objects.equals(params, postLink.params);
+    }
+
+    @Override
+    public int hashCode() {
+
```
oswaldquek: empty line

src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private Client client;
+
+    @Mock
+    private PublicApiConfig configuration;
+    private PublicApiUriGenerator publicApiUriGenerator;
+    private ConnectorUriGenerator connectorUriGenerator;
+    
+    @Before
+    public void setup() {
```
oswaldquek: not sure the comments in this setup() method are needed. personally i think they're self explanatory...

src/main/java/uk/gov/pay/api/model/links/Link.java
alexbishop1: It’s what IntelliJ automatically generated

src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
alexbishop1: The intention was to explain what’s going on to Pact newbies

src/main/java/uk/gov/pay/api/model/links/Link.java
oswaldquek: ok then, in that case :)

src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
oswaldquek: i still think `// These can be concrete implementations, because they're simple` isn't really needed...

src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
```diff
@@ -0,0 +1,82 @@
...
+        when(configuration.getConnectorUrl()).thenReturn(connectorRule.getUrl()); // We will actually send real requests here, which will be intercepted by pact
+
+        when(configuration.getBaseUrl()).thenReturn("http://publicapi.test.localhost/"); // Has to match what is in the Pact
```
oswaldquek: i'm also not sure what this comment means and how it relates to pact. if anything, the publicapi-connector.json pact file has a "self" link `http://connector.service.backend/v1/api/accounts/GATEWAY_ACCOUNT_ID/charges/ch_ab2341da231434l`.

oswaldquek: i think it's actually related to publicApiUriGenerator rather than pact

src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
alexbishop1: Yeah, I wasn’t too keen on that. I’ll remove.

PR #189, opened 2018-06-15
Bau add pact testing

georgeracu:
## WHAT
Added Pac test for `getPayment` for cards

## HOW 
* Added separate json files for CARD and DIRECT_DEBIT tokens as Pact gets confused is using the
same path for different interactions. Next step is to set the state in Povider in order to merge these
two files
* Added `publicapi-connector` json file for getting a charge from connector
* Extracted common field into a base class, so it can be reused
* Added directories for each payment type, for easier navigation around


heathd:
I'd strongly prefer to write the pact test in the style shown in #186. 

Let's get the tests passing on that PR and then follow the same pattern here.

georgeracu:
Replaced by other branch

PR #193, opened 2018-06-19
PP-3678 Split PaymentsResourceITest

heathd:
separate files for `CreatePaymentITest` and `GetPaymentITest`

with @alexbishop1


src/test/java/uk/gov/pay/api/it/CreatePaymentITest.java
```diff
@@ -0,0 +1,233 @@
...
+
+    private static final ZonedDateTime TIMESTAMP = DateTimeUtils.toUTCZonedDateTime("2016-01-01T12:00:00Z").get();
+    private static final int AMOUNT = 9999999;
+    private static final String CHARGE_ID = "ch_ab2341da231434l";
+    private static final String CHARGE_TOKEN_ID = "token_1234567asdf";
+    private static final PaymentState CREATED = new PaymentState("created", false, null, null);
+    private static final RefundSummary REFUND_SUMMARY = new RefundSummary("pending", 100L, 50L);
+    private static final String PAYMENT_PROVIDER = "Sandbox";
+    private static final String CARD_BRAND_LABEL = "Mastercard";
+    private static final String RETURN_URL = "https://somewhere.gov.uk/rainbow/1";
```
simad: We should probably use a `.test` url as @alexbishop1 would suggest

alexbishop1: I’m actually OK with `*.something.gov.uk` domains because we’re the government and it’s unlikely to result in grief or embarrassment for anyone.

It’s other random potentially-registered domains I worry about.

PR #195, opened 2018-06-20
PP-3679 Add DD search by agreement to PaymentResource

SandorArpa:
- add direct debit search to PaymentSearchService
- add search Response process
- add Pact test


SandorArpa:
Needs refactoring

PR #197, opened 2018-06-21
PP-3677 Add AgreementError domain object

simad:
## WHAT
To be used in Agreement-related api calls. Most of the errors might overlap with card payments, but we still keep the objects separate to allow us to diverge at will.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/directdebit/agreement/AgreementError.java
+++ b/src/main/java/uk/gov/pay/api/model/directdebit/agreement/AgreementError.java
@@ -1,0 +1,90 @@
+package uk.gov.pay.api.model.directdebit.agreement;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+import static com.google.common.collect.ObjectArrays.concat;
+import static java.lang.String.format;
+
+@ApiModel(value = "AgreementError", description = "An Agreement Error response")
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/api/model/directdebit/agreement/AgreementError.java
```diff
@@ -0,0 +1,90 @@
...
+        this.description = format(code.getFormat(), parameters);
+    }
+
+    private AgreementError(String fieldName, Code code, Object... parameters) {
+        this.field = fieldName;
+        this.code = code;
+        this.description = format(code.getFormat(), concat(fieldName, parameters));
+    }
+
+    @ApiModelProperty(example = "amount")
```
DanailMinchev: Currently we don't have `amount` in the API, should an example be something from agreement API?

src/main/java/uk/gov/pay/api/model/directdebit/agreement/AgreementError.java
```diff
@@ -0,0 +1,90 @@
...
+    public String getField() {
+        return field;
+    }
+
+    @ApiModelProperty(example = "P0102")
+    public String getCode() {
+        return code.value();
+    }
+
+    @ApiModelProperty(example = "Invalid attribute value: amount. Must be less than or equal to 10000000")
```
DanailMinchev: If you change the field example please update this description as well

src/main/java/uk/gov/pay/api/model/directdebit/agreement/AgreementError.java
simad: 👍 well spotted, I skim over annotations :D

PR #199, opened 2018-06-21
bau: Quick fix for failing gitBranchName method

oswaldquek:
gitBranchName is currently not returning 'master' on master branch builds. This
is currently blocking progress so doing this quick fix now.

@oswaldquek

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,11 +25,14 @@ pipeline {
 
   stages {
     stage('Maven Build') {
+      when {
+        branch 'master'
+      }
       steps {
         script {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -47,6 +50,34 @@ pipeline {
...
+          }
+          postSuccessfulMetrics("publicapi.maven-build", stepBuildTime)
+      }
+      }
+      post {
+          failure {
+              postMetric("publicapi.maven-build.failure", 1)
+          }
+      }
+    }
```
mrlumbu: you could have just conditionally set the branch name rather than repeat everything no?

PR #200, opened 2018-06-22
PP-3679 Add DD search by agreement to PaymentResource

SandorArpa:
Refactor `PaymentsResource` to use `PaymentSearchService`. This is needed so we can implement `Pact` test(s).
Implement Direct Debit search. Use pact testing to test the DD search instead of mocks.
Search is based on `Account` type. This is needed as the nature of the backend response is differ for connector and direct-debit-connector. Instead of trying to make existing `PaymentForSearchResult` generic we opted to separate the searches and thus introduced `DDSearchResponse`.

- introduce `IPaymentSearchPagination` for simplified pagination building.
- add dd search to PaymentSearchService
- add relevant PoJos (`DDSearchResponse`, `DDTransactionFromResponse`,  `DDTransactionForSearch`, etc)
- add Response process from back end responses
- add Pact test

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/IPaymentSearchPagination.java
+++ b/src/main/java/uk/gov/pay/api/model/IPaymentSearchPagination.java
@@ -1,0 +1,10 @@
+package uk.gov.pay.api.model;
+
+import uk.gov.pay.api.model.links.PaymentSearchNavigationLinks;
+
+public interface IPaymentSearchPagination {
+    int getCount();
+    int getTotal();
+    int getPage();
+    PaymentSearchNavigationLinks getLinks();
+}
--- a/src/main/java/uk/gov/pay/api/model/directdebit/DDPaymentState.java
+++ b/src/main/java/uk/gov/pay/api/model/directdebit/DDPaymentState.java
@@ -1,0 +1,68 @@
+package uk.gov.pay.api.model.directdebit;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import java.util.Objects;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@ApiModel(value="DDPaymentState", description = "A structure representing the current state of the direct debit transaction in its lifecycle.")
+public class DDPaymentState {
+    @JsonProperty("status")
+    private String status;
+
+    @JsonProperty("finished")
+    private boolean finished;
+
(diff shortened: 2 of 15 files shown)


src/main/java/uk/gov/pay/api/model/IPaymentSearchPagination.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.pay.api.model;
+
+import uk.gov.pay.api.model.links.PaymentSearchNavigationLinks;
+
+public interface IPaymentSearchPagination {
```
DanailMinchev: 👍

src/main/java/uk/gov/pay/api/model/directdebit/DDPaymentState.java
```diff
@@ -0,0 +1,68 @@
...
+    private boolean finished;
+
+    public static DDPaymentState createPaymentState(JsonNode node) {
+        return new DDPaymentState(
+            node.get("status").asText(),
+            node.get("finished").asBoolean()
+        );
+    }
+
+    public DDPaymentState() {
```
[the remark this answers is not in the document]
SandorArpa: Not sure, I just stripped down `PaymentState` to fit our purpose

src/test/resources/pacts/publicapi-direct-debit-connector-search-by-mandate-three-results.json
```diff
@@ -0,0 +1,269 @@
...
+        "body": {
+          "total": 3,
+          "count": 3,
+          "page": 1,
+          "results": [
+            {
+              "amount": 1234,
+              "transaction_id": "askd87sfnsadn489nfd8ddnv",
+              "name": "J. Citizen",
+              "email": "j.citizen@mail.fake",
```
DanailMinchev: I see we already use `.fake` in our code, we updated some of it to use `*example.com` or `*.test`, so may be it's good to use test tlds epecially for emails?

src/main/java/uk/gov/pay/api/model/directdebit/search/DDSearchResponse.java
```diff
@@ -0,0 +1,45 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.pay.api.model.IPaymentSearchPagination;
+import uk.gov.pay.api.model.links.PaymentSearchNavigationLinks;
+
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class DDSearchResponse implements IPaymentSearchPagination {
```
DanailMinchev: Not sure about `DD` prefix as name, we usually use `DirectDebit`. I understand that it may be too long, so I don't mind, if you would like please check with someone else as well

SandorArpa: It's under package `directdebit` so it should be self explanatory why we use DD (to avoid mile long class names)

DanailMinchev:
Actually, just tried on my machine: `swagger.json` was not regenerated, so no need to commit it.

PR #201, opened 2018-06-22
PP-3845: Don't publish pacts for which there are no provider tests

oswaldquek:
PP-3845 is about checking pact compatibility between publicapi and
direct-debit-connector; however we currently also publish pacts for publicauth
and connector, and these have no provider tests written for them yet, and
therefore no pact verifications. We need to disable publishing them so the
pact-broker has no knowledge of them and can check pact compatibility between
publicapi and direct-debit-connector only.

@oswaldquek

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,7 +104,7 @@
         <dependency>
             <groupId>uk.gov.pay</groupId>
             <artifactId>testing</artifactId>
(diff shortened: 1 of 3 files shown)


tlwr:
I approve of the ordering of prepositions in this PR title

PR #203, opened 2018-06-22
PP-3677 Publicapi should return a public self url for agreements

simad:
## WHAT
 - We were exposing connector's internal url, which is not exactly the right thing to do


src/main/java/uk/gov/pay/api/resources/AgreementsResource.java
```diff
@@ -39,7 +39,7 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(AgreementsResource.class);
 
     private final String baseUrl;
-    private AgreementService agreementService;
+    private final AgreementService agreementService;
```
mrlumbu: I know it's not your fault, but `import uk.gov.pay.api.model.links.PaymentWithAllLinks;
` is unused.

src/main/java/uk/gov/pay/api/service/AgreementService.java
```diff
@@ -53,17 +55,24 @@ public GetAgreementResponse get(Account account, String agreementId) {
...
         }
         throw new GetAgreementException(connectorResponse);
     }
 
+    private AgreementLinks createLinksFromMandateResponse(MandateConnectorResponse mandate) {
+        AgreementLinks agreementLinks = new AgreementLinks();
+        agreementLinks.addSelf(publicApiUriGenerator.getAgreementURI(mandate.getMandateId()).toString());
+        agreementLinks.addKnownLinksValueOf(mandate.getLinks());
+        return agreementLinks;
+    }
```
mrlumbu: new line

PR #204, opened 2018-06-26
PP-3678 Add collect payment from agreement functionality to public API

alexbishop1:
Add support for an `agreement_id` field in the payload to create a payment. If such a field is present, the request becomes a request to collect from an existing agreement (which initially will always be a Direct Debit mandate).

If an `agreement_id` is specified, the `return_url` is not required and will be ignored (because collecting from an agreement does not require any input from the paying user).

The Pact test is incomplete; it will be finished after we complete the Direct Debit connector work.

with @simad


src/test/java/uk/gov/pay/api/it/validation/PaymentsResourceAgreementIdValidationITest.java
```diff
@@ -0,0 +1,197 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+import static com.jayway.restassured.RestAssured.given;
+import static com.jayway.restassured.http.ContentType.JSON;
+import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.core.Is.is;
+
+public class PaymentsResourceAgreementIdValidationITest extends PaymentResourceITestBase {
```
DanailMinchev: 👍 good test coverage

src/test/java/uk/gov/pay/api/it/validation/PaymentsResourceAmountValidationITest.java
```diff
@@ -28,7 +29,7 @@ public void createPayment_responseWith422_whenAmountIsNegative() throws IOExcept
                 "  \"amount\" : -123," +
                 "  \"reference\" : \"Some reference\"," +
                 "  \"description\" : \"Some description\"," +
-                "  \"return_url\" : \"http://somewhere.gov.uk/rainbow/1\"" +
+                "  \"return_url\" : \"https://somewhere.gov.uk/rainbow/1\"" +
```
DanailMinchev: It was already like this, but probably we can use example.com, because we updated the docs as well - https://github.com/alphagov/pay-publicapi/commit/459919fbfc93bde7b5f4a9b286d85a9b2e5e7f98

src/test/java/uk/gov/pay/api/it/validation/PaymentsResourceDescriptionValidationITest.java
```diff
@@ -29,7 +30,7 @@ public void createPayment_responseWith400_whenDescriptionIsNumeric() throws IOEx
                 "  \"amount\" : 9900," +
                 "  \"description\" : 1234," +
                 "  \"reference\" : \"Some reference\"," +
-                "  \"return_url\" : \"http://my-payments.com\"" +
+                "  \"return_url\" : \"https://example.com\"" +
```
DanailMinchev: 👍

src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
```diff
@@ -42,31 +42,50 @@ public void setup() {
...
+    
     @Test
     public void deserialize_shouldThrowBadRequestException_whenJsonIsNotWellFormed() throws Exception {
 
         String invalidJson = "{" +
                 "  \"amount\" : " +
                 "  \"reference\" : \"Some reference\"," +
                 "  \"description\" : \"Some description\"," +
-                "  \"return_url\" : \"http://somewhere.gov.uk/rainbow/1\"" +
+                "  \"return_url\" : \"https://somewhere.gov.uk/rainbow/1\"" +
```
DanailMinchev: same as above

src/test/java/uk/gov/pay/api/it/validation/PaymentsResourceAmountValidationITest.java
alexbishop1: Meh, it’s only a test, doesn’t necessarily have to match the docs

I’m OK with using `something.gov.uk` domains where it probably would be a `gov.uk` domain (it’s an exception to my only-use-domain-names-reserved-in-RFC-2606-and-RFC-6761 rule)

PR #206, opened 2018-06-27
Gateway account is numeric

mrlumbu:


mrlumbu:
@alexbishop1 https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/util/DatabaseTestHelper.java#L42

PR #224, opened 2018-07-13
PP-4005 Rate filter audit

mrlumbu:
Added new method 'auditRateOf' to RateLimiterFilter to log no of transactions/second for service. This will help to find out optimal rate against current rate of `25tx/s` for each node .  

Does not effect current functionality but does additional logging if `tx/s > 4`

We have chosen `4 tx/s` from inspecting logs to see what kind of load a single service generates. MTP can generate up to about `7/8 tx/s`, but since this will be spread over 3 nodes, the 'normal' rate limit per service will be `12 tx/s`, which should be fine. However there may be cases where the effective limit becomes equal to the limit per node (i.e. if all requests for a particular service were directed at one node). Therefore the purpose of auditing the limit in this manner is to establish whether, and how often, a limit of `4 tx/s` per node would affect MTP (or any other service).

with @kbottla

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/RateLimiter.java
+++ b/src/main/java/uk/gov/pay/api/filter/RateLimiter.java
@@ -46,8 +46,7 @@ public class RateLimiter {
             auditCache.get(key, () -> new RateLimit(auditRate, perMillis)).updateAllowance();
         } catch (RateLimitException e) {
             LOGGER.info(String.format(
-                    "Rate limit reached for rate limit key %s using rate of %d requests per %d ms",
-                    key,
+                    "Rate limit reached for a service using rate of %d requests per %d ms",
                     auditRate,
                     perMillis)
             );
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/api/filter/RateLimiter.java
```diff
@@ -33,4 +40,21 @@ void checkRateOf(String key) throws RateLimitException {
             LOGGER.error("Unexpected error creating a Rate Limiter object in cache", e);
         }
     }
+
+    void auditRateOf(String key) {
+        try {
+            auditCache.get(key, () -> new RateLimit(auditRate, perMillis)).updateAllowance();
+        } catch (RateLimitException e) {
+            LOGGER.info(String.format(
+                    "Rate limit reached for rate limit key %s using rate of %d requests per %d ms",
```
simad: won't this log bearer tokens? ❓

mrlumbu: uh yeah

mrlumbu: 😬

kbottla: Removed bearer token from log.

PR #233, opened 2018-07-27
PP-4005 Segment rate limit by method

mrlumbu:
In order to more precisely rate limit the different parts of
public api, this commit passes a concatenation of method and
api key as rate limiting key. This will mean GET requests cannot impinge
on POSTs, so if a service does a batch of GETs it won't block the
mor eimportant process of creating new payments. This is a small
change in right direction, rather than a total solution.

We may want to set different rates for POSTs and GETs at some point,
but this is not currently required.

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
+++ b/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
@@ -53,14 +53,14 @@ public class RateLimiterFilter implements Filter {
         final String method = ((HttpServletRequest) request).getMethod();
 
         rateLimiter.auditRateOf(method + "-" + authorization);
-        
+
         try {
-            rateLimiter.checkRateOf(authorization);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
```diff
@@ -53,14 +53,14 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+
         try {
-            rateLimiter.checkRateOf(authorization);
+            rateLimiter.checkRateOf(method + "-" + authorization);
             chain.doFilter(request, response);
         } catch (RateLimitException e) {
             LOGGER.info("Rate limit reached for current service. Sending response '429 Too Many Requests'");
             setTooManyRequestsError((HttpServletResponse) response);
-        }
+        }
```
rhowe-gds: sneaky space

PR #237, opened 2018-08-01
PP-3680 Adding validation for display_size

danworth:
- Added validation for display_size, future work may want to refactor validation
into a common reusable class. Also may be prudent to refactor ValidationException to be less payment specific in future.
- Separated serialization and deserialization for mandateId and transactionId
on DDEvent to permit correct setting of names in json.
- Created and updated tests to check for correct naming convention on DDEvent and validation.

### DIFF
--- a/src/test/java/uk/gov/pay/api/validation/DirectDebitEventSearchValidatorTest.java
+++ b/src/test/java/uk/gov/pay/api/validation/DirectDebitEventSearchValidatorTest.java
@@ -1,0 +1,48 @@
+package uk.gov.pay.api.validation;
+
+
+import org.junit.Test;
+import uk.gov.pay.api.exception.ValidationException;
+
+import java.time.ZonedDateTime;
+
+import static org.junit.Assert.assertTrue;
+
+public class DirectDebitEventSearchValidatorTest {
+    
+    
+    @Test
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/api/validation/DirectDebitEventSearchValidatorTest.java
```diff
@@ -0,0 +1,48 @@
...
+        DirectDebitEventSearchValidator.validateSearchParameters(ZonedDateTime.now().toString(), ZonedDateTime.now().toString(), 1);
+    }
+
+    @Test(expected = ValidationException.class)
+    public void shouldFailWithInvalidDisplaySize() {
+        try {
+            DirectDebitEventSearchValidator.validateSearchParameters(ZonedDateTime.now().toString(), ZonedDateTime.now().toString(), 501);
+        } catch (ValidationException ex) {
+            assertTrue(ex.getPaymentError().getDescription().contains("display_size"));
+            throw ex;
```
alexbishop1: Another way to do this would be to remove the `expected = ValidationException.class`, not rethrow the exception, and add a `fail()` call after the `catch` block. But this works!

PR #240, opened 2018-08-02
PP 4005 - Separate rate limits for GET and POST

kbottla:
Changes to set different rate limits for GET and POST methods. 

GET requests : revert back to 25 requests/second/node
POST requests : keep current limit of 5 request/second/node


kbottla:
Thanks @danworth

PR #241, opened 2018-08-08
PP_4036 - Ratelimiting using Redis

kbottla:
Redis based solution to support rate limit API requests. Enhances current solution so that public api nodes can lookup redis for accurate rate limits instead of local cache. When redis is not available the solution will fall back to current solution (using local cache) and logs errors appropriately.

Changes include: 
 - Redis for RateLimiting
 - RateLimit test with Jedis library using Redis
 - Added docker-client library for Integration tests
 - Fixed issue with failing integration tests
 - Fallback to local rate limit solution when Redis is unavailable & tests


oswaldquek:
looking...

src/main/java/uk/gov/pay/api/app/config/PublicApiModule.java
```diff
@@ -64,9 +66,26 @@ public ObjectMapper provideObjectMapper() {
...
+
+        return new RateLimiter(configuration.getRateLimiterConfig().getRate(),
                 configuration.getRateLimiterConfig().getRateForPost(),
-                configuration.getRateLimiterConfig().getAuditRate(), 
-                configuration.getRateLimiterConfig().getPerMillis());
+                configuration.getRateLimiterConfig().getAuditRate(),
+                configuration.getRateLimiterConfig().getPerMillis(),
+                configuration.getRateLimiterConfig().getNoOfNodes(),
+                localRateLimiter,
+                getRedisRateLimiter());
```
oswaldquek: this should use the redisRateLimiter variable in line 71

src/main/java/uk/gov/pay/api/app/config/PublicApiModule.java
```diff
@@ -64,9 +66,26 @@ public ObjectMapper provideObjectMapper() {
 
     @Provides
     public RateLimiter provideRateLimiter() {
-        return new RateLimiter(configuration.getRateLimiterConfig().getRate(), 
+
+        LocalRateLimiter localRateLimiter=getLocalRateLimiter();
+        RedisRateLimiter redisRateLimiter=getRedisRateLimiter();
```
oswaldquek: spacings.
`localRateLimiter = getLocalRateLimiter()` 
`redisRateLimiter = getRedisRateLimiter()`

src/main/java/uk/gov/pay/api/filter/ratelimit/LocalRateLimiter.java
```diff
@@ -0,0 +1,34 @@
...
+                .expireAfterAccess(perMillis, TimeUnit.MILLISECONDS)
+                .build();
+    }
+
+    void checkRateOf(String key, int rate) throws RateLimitException {
+        try {
+            cache.get(key, () -> new RateLimit(rate, perMillis)).updateAllowance();
+        } catch (ExecutionException e) {
+            //ExecutionException is thrown when the valueLoader throws a checked exception.
+            //We just create a new instance so no exceptions will be thrown, this should never happen.
```
oswaldquek: I don't understand this comment. what's `valueLoader`? and how are we creating a new instance? I suggest some re-wording here.

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -12,50 +13,51 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RateLimiter.class);
 
-    private final int rate;
-    private final int rateForPost;
+    private int rate;
+    private int rateForPost;
+    private int auditRate;
+    private int perMillis;
+    private int noOfNodes;
```
oswaldquek: above fields should be final

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -12,50 +13,51 @@
...
     private static final String METHOD_POST = "POST";
     
-    private final int auditRate;
-    private final int perMillis;
-    private final Cache<String, RateLimit> cache;
     private final Cache<String, RateLimit> auditCache;
-
-    public RateLimiter(int rate, int rateForPost, int auditRate, int perMillis) {
+    LocalRateLimiter localRateLimiter;
+    RedisRateLimiter redisRateLimiter;
```
oswaldquek: is there a reason why redisRateLimiter and localRateLimiter aren't private?

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -12,50 +13,51 @@
...
-        } catch (ExecutionException e) {
-            //ExecutionException is thrown when the valueLoader throws a checked exception.
-            //We just create a new instance so no exceptions will be thrown, this should never happen.
-            LOGGER.error("Unexpected error creating a Rate Limiter object in cache", e);
+            redisRateLimiter.checkRateOf(key, getRateForMethod(method), perMillis);
+        } catch (RedisRateLimitException e) {
+            LOGGER.warn("Exception occurred checking rate limits using RedisRateLimiter, falling back to LocalRateLimiter");
+            
+            /* Used as fail safe mechanism in case redis is unavailable. 
+               Local rate Limiting should not replace redis for accurate rate limiting */
```
oswaldquek: i don't think this comment adds value as it's obvious we're falling back to local rate limiting here

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -12,50 +13,51 @@
...
 
-    private final int rate;
-    private final int rateForPost;
+    private int rate;
+    private int rateForPost;
+    private int auditRate;
+    private int perMillis;
+    private int noOfNodes;
     
     private static final String METHOD_POST = "POST";
```
oswaldquek: "POST" is already defined in javax.ws.rs.HttpMethod.POST, can we use that please.

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -68,4 +70,21 @@ void auditRateOf(String key) {
             LOGGER.error("Unexpected error creating a Rate Limiter object in cache", e);
         }
     }
+
+    private int getRateForMethod(String method){
+        if(METHOD_POST.equals(method)){
```
oswaldquek: use `javax.ws.rs.HttpMethod.POST.equals(method)`

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -68,4 +70,21 @@ void auditRateOf(String key) {
...
+        }
+        return rate;
+    }
+
+    /**
+     * Returns rate to use local cache. It is the rate divided by no. of publicapi nodes
+     * 
+     * @param method GET, POST ...
+     * @return
+     */
```
oswaldquek: no need for javadoc here as the method itself is self documenting

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -68,4 +70,21 @@ void auditRateOf(String key) {
...
+    private int getLocalRateForMethod(String method){
+       return getRateForMethod(method)/noOfNodes;
```
oswaldquek: what if this division doesn't divide nicely to an integer? does the rate round up or down to the nearest integer? be good to add tests for it

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -0,0 +1,103 @@
...
+
+        if (count == 1) {
+            getResource().expire(derivedKey, perMillis/1000);
+        }
+
+        return count;
+    }
+
+    private Jedis getResource() {
+        return jedisPool.getResource();
```
oswaldquek: `jedisPool.getResource()` throws redis.clients.jedis.exceptions.JedisException and redis.clients.jedis.exceptions.JedisConnectionException so we should catch these exceptions in the calling methods

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -12,50 +13,51 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RateLimiter.class);
 
-    private final int rate;
-    private final int rateForPost;
+    private int rate;
+    private int rateForPost;
+    private int auditRate;
+    private int perMillis;
```
oswaldquek: is this requests per millis? it would be better to be more explicit, e.g. `requestPerMillis`

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -0,0 +1,103 @@
...
+        this.jedisPool = jedisPool;
+    }
+
+    /**
+     * This block needs to be synchronous. Each RateLimit object will be shared between requests
+     * from the same source (Service), so is not shared across all the requests.
+     *
+     * @throws RateLimitException
+     */
+    synchronized void checkRateOf(String key, int rate, int perMillis)
```
oswaldquek: consider renaming `perMillis` to `requestPerMillis`

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -0,0 +1,103 @@
...
+            throw new RedisRateLimitException();
+        }
+
+        if (count!=null && count > rate) {
+            LOGGER.info(String.format("rate exceeded - count: %d, rate: %d", count, rate));
+            throw new RateLimitException();
+        }
+    }
+
+    synchronized private Long updateAllowance(String key, int perMillis) {
```
oswaldquek: consider renaming `perMillis` to `requestPerMillis`

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -0,0 +1,103 @@
...
+     * <p>
+     * - perMillis < 1000 : Window calculated for milliseconds    
+     * - perMillis >=1000 && <=60000 : Window calculated for seconds(s)
+     * - perMillis >60000 && <=3600000 : Window calculated for minute(s)
+     * - perMillis >3600000 && <=86400000 : Window calculated for hours(s)
+     *
+     * @param perMillis
+     * @return new key based on perMillis (works for second/minute/hour windows only)
+     */
+    private String getKeyForWindow(String key, int perMillis) throws OutOfScopeException {
```
oswaldquek: consider renaming `perMillis` to `requestPerMillis`

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -0,0 +1,103 @@
...
+
+        LocalDateTime now = LocalDateTime.now();
+
+        int slot;
```
oswaldquek: rename `slot` to `window` for consistency, since you're talking about "Window" in the javadoc

src/main/resources/config/config.yaml
```diff
@@ -27,10 +27,19 @@ jerseyClientConfig:
   disabledSecureConnection: ${DISABLE_INTERNAL_HTTPS}
 
 rateLimiter:
-  rate: ${RATE_LIMITER_VALUE:-25}  # the rate is for all requests except POST
-  rateForPost: ${RATE_LIMITER_VALUE_POST:-5}
+  rate: ${RATE_LIMITER_VALUE:-75}  # the rate is for all requests except POST
```
oswaldquek: is this rate for ALL the nodes? if so it seems wrong. i don't think the service should have information on its environment, in this case the number of nodes. what if we do autoscaling? what if we change desired nodes in aws manually? i think as far as this config is concerned, all it should care about is the rate for ONE node.

oswaldquek:
As a general comment, I don't see the benefit of having the unit tests LocalRateLimiterTest, RateLimiterTest and RedisRateLimiterTest. It seems like they're just adding a lot of noise. I'm not a fan of excessive mocking either as they make tests difficult to read. I think rate limiting is a good candidate for integration-like tests (I say integration-like as it's not really integration in the sense of spinning up the entire application, but I mean using concrete classes instead of mocking things). Basically I want to know three simple things:

1. does the rate limit work for a given key?
2. does the rate limit work for multiple keys? as in, is the rate limiter clever enough to have separate rate limits for different keys.
3. does the local rate limit kick in if redis isn't available?

The above would be fairly simple integration tests where we send in a mocked HttpServletRequests.

I would suggest a LocalRateLimiterTest which tests questions 1 and 2 using real classes.
Then a RedisRateLimiterTest which tests questions 1, 2 and 3 using real classes.

Also if we get rid of the knowledge of number of nodes in the config.yaml it will make code and testing simpler too.

src/test/java/uk/gov/pay/api/it/ResourcesFiltersLocalRateLimiterITest.java
```diff
@@ -36,7 +36,7 @@
 import static uk.gov.pay.api.it.fixtures.PaginatedPaymentSearchResultFixture.aPaginatedPaymentSearchResult;
 import static uk.gov.pay.api.it.fixtures.PaymentSearchResultBuilder.aSuccessfulSearchPayment;
 
-public class ResourcesFiltersITest extends PaymentResourceITestBase {
+public class ResourcesFiltersLocalRateLimiterITest extends PaymentResourceITestBase {
```
oswaldquek: Looks like PaymentResourceITestBase is using a real redis url, so how is this test testing local rate limiting?

src/test/java/uk/gov/pay/api/it/ResourcesFiltersRedisRateLimiterITest.java
```diff
@@ -0,0 +1,330 @@
...
+import static com.jayway.restassured.RestAssured.given;
+import static com.jayway.restassured.http.ContentType.JSON;
+import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
+import static org.hamcrest.core.IsCollectionContaining.hasItem;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+import static uk.gov.pay.api.it.fixtures.PaginatedPaymentSearchResultFixture.aPaginatedPaymentSearchResult;
+import static uk.gov.pay.api.it.fixtures.PaymentSearchResultBuilder.aSuccessfulSearchPayment;
+
+public class ResourcesFiltersRedisRateLimiterITest extends PaymentResourceITestBase {
```
oswaldquek: given RateLimiterFilter determines if a request is a GET, i think testing all the get methods, and testing the rate limit and invalid authorization header for each method is probably overkill. i think some tests could be removed from here.

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * This block needs to be synchronous. Each RateLimit object will be shared between requests
+     * from the same source (Service), so is not shared across all the requests.
+     *
+     * @throws RateLimitException
+     */
+    synchronized void checkRateOf(String key, int rate, int perMillis)
+            throws RedisRateLimitException, RateLimitException {
```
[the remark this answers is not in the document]
kbottla: Possible exceptions here are JedisException, JedisConnectionException , OutOfScopeException, NoSuchElementException and a General Exception. I believe new exception will be clean to handle in calling methods ?

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimitException.java
```diff
@@ -0,0 +1,3 @@
+package uk.gov.pay.api.filter.ratelimit;
+
+public class RedisRateLimitException extends Exception {}
```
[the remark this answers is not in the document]
kbottla: Possible exceptions here are JedisException, JedisConnectionException , OutOfScopeException, NoSuchElementException and a General Exception. I believe new exception will be clean to handle in calling methods ?

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
kbottla: Possible exceptions here are JedisException, JedisConnectionException , OutOfScopeException, NoSuchElementException and a General Exception. I believe new exception will be clean to handle in calling methods ?

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+    private volatile boolean stopped=false;
+
+    private static final int DB_TIMEOUT_SEC=10;
+    private static final String REDIS_IMAGE="redis:latest";
+    private static final String INTERNAL_PORT="6379";
+
+    public RedisContainer(DockerClient docker, String host) throws DockerException, InterruptedException, IOException, ClassNotFoundException {
+
+        this.docker=docker;
+        this.host=host;
```
oswaldquek: spaces. `this.host = host` and similar for docker line above

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+
+        failsafeDockerPull(docker, REDIS_IMAGE);
+        docker.listImages(DockerClient.ListImagesParam.create("name", REDIS_IMAGE));
+
+        final HostConfig hostConfig = HostConfig.builder().logConfig(LogConfig.create("json-file")).publishAllPorts(true).build();
+        ContainerConfig containerConfig=ContainerConfig.builder()
+                .image(REDIS_IMAGE)
+                .hostConfig(hostConfig)
+                .build();
+        containerId=docker.createContainer(containerConfig).id();
```
oswaldquek: see comment on spaces

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+        docker.startContainer(containerId);
+        port=hostPortNumber(docker.inspectContainer(containerId));
```
oswaldquek: see comment on spaces

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+            logger.error("Docker image "+image+" could not be pulled from DockerHub", e);
+        }
+    }
+
+    public String getConnectionUrl() {
+        return host+":"+port;
+    }
+
+    private static int hostPortNumber(ContainerInfo containerInfo) {
+        List<PortBinding> portBindings=
```
oswaldquek: see comment on spaces

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
kbottla: It is rate per millis.

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+    private void failsafeDockerPull(DockerClient docker, String image) {
+        try {
+            docker.pull(image);
+        } catch (Exception e) {
+            logger.error("Docker image "+image+" could not be pulled from DockerHub", e);
+        }
+    }
+
+    public String getConnectionUrl() {
+        return host+":"+port;
```
oswaldquek: spaces. `host + ":" + port`

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+        logger.info("Redis host port: {}", portBindings.stream().map(PortBinding::hostPort).collect(joining(", ")));
+        return parseInt(portBindings.get(0).hostPort());
+    }
+
+    private void registerShutdownHook() {
+        Runtime.getRuntime().addShutdownHook(new Thread(this::stop));
+    }
+
+    private void waitForRedisToStart() throws DockerException, InterruptedException, IOException {
+        Stopwatch timer=Stopwatch.createStarted();
```
oswaldquek: see comment on spaces

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
kbottla: It is rate per millis.

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+        boolean succeeded=false;
```
oswaldquek: see comment on spaces

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
kbottla: It is rate per millis.

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
kbottla: It is rate per millis.

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+            succeeded=checkRedisConnection();
+        }
+        if (!succeeded) {
+            throw new RuntimeException("Redis did not start in "+DB_TIMEOUT_SEC+" seconds.");
+        }
+        logger.info("Redis docker container started in {}.", timer.elapsed(TimeUnit.MILLISECONDS));
+    }
+
+    private boolean checkRedisConnection() throws IOException {
+        try (Jedis jedis=new Jedis(host, port)) {
```
oswaldquek: see comment on spaces

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+            return false;
+        }
+    }
+
+    public void stop() {
+        if (stopped) {
+            return;
+        }
+        try {
+            stopped=true;
```
oswaldquek: see comment on spaces

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+            System.err.println("Killing redis container with ID: "+containerId);
+            LogStream logs=docker.logs(containerId, DockerClient.LogsParam.stdout(), DockerClient.LogsParam.stderr());
```
oswaldquek: see comment on spaces

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -0,0 +1,118 @@
...
+            System.err.println("Killed container logs:\n");
+            logs.attach(System.err, System.err);
+            docker.stopContainer(containerId, 5);
+            docker.removeContainer(containerId);
+        } catch (DockerException | InterruptedException | IOException e) {
+            System.err.println("Could not shutdown "+containerId);
```
oswaldquek: see comment on spaces

src/test/java/uk/gov/pay/api/it/rule/RedisDockerRule.java
```diff
@@ -0,0 +1,65 @@
...
+import com.spotify.docker.client.exceptions.DockerException;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+public class RedisDockerRule implements TestRule {
```
oswaldquek: fix all spacing issues during an assignment

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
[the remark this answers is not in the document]
kbottla: NoSuchElementException is thrown here https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/util/Pool.java#L51 and borrowObject raises general Exception. We will risk ending up with unhandled runtime exception if we catch specific ones only.

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
kbottla: we always set rate as millis and usually set to 1000 ~ 1 second. So it will effectively be "x requests per second". This allows us to reduce the granularity to milli seconds if required.

src/test/java/uk/gov/pay/api/it/rule/RedisDockerRule.java
kbottla: It is a shame I overlooked it. I have updated IDE code style for spaces. So should fix it for future changes.

src/main/java/uk/gov/pay/api/app/config/PublicApiModule.java
kbottla: Updated

src/main/java/uk/gov/pay/api/app/config/PublicApiModule.java
kbottla: Updated..

src/main/java/uk/gov/pay/api/filter/ratelimit/LocalRateLimiter.java
kbottla: Updated comment for value loader and new instance

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
kbottla: Updated..

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
kbottla: Changed to private..

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
kbottla: Agree. Removed comment

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
kbottla: Thank you, changed to HttpMethod.POST

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
kbottla: changed to HttpMethod.POST

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
kbottla: Remove javadoc

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
kbottla: Not applicable anymore. Removed noOfNodes and added new environment variables to use for local rate limiting

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -0,0 +1,103 @@
...
+    private JedisPool jedisPool;
+
+    @Inject
+    public RedisRateLimiter(JedisPool jedisPool) {
+        this.jedisPool = jedisPool;
+    }
+
+    /**
+     * This block needs to be synchronous. Each RateLimit object will be shared between requests
+     * from the same source (Service), so is not shared across all the requests.
```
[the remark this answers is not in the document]
kbottla: This comment is unnecessary for redis, removed comment.

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
kbottla: Renamed..

src/main/resources/config/config.yaml
kbottla: I was thinking to save on adding more environment variables. I guess this is making it worse. So added new env variable to use for local rate limits. This will give us flexibility to apply local rates using pay infra.

src/test/java/uk/gov/pay/api/it/ResourcesFiltersLocalRateLimiterITest.java
kbottla: This was cause of moving RedisDockerRule to Base class. I have removed Base class dependency for ResourcesFiltersLocalRateLimiterITest and

src/test/java/uk/gov/pay/api/it/ResourcesFiltersRedisRateLimiterITest.java
kbottla: I think renaming files caused confusion here. I have restored ResourcesFiltersITest which is testing all public API endpoints for different errors returned. I now created only one class ResourcesFiltersLocalRateLimiterITest which do not access redis and few tests around it. All other tests use Redis by default.

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
kbottla: Done..

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
kbottla: Done..

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
kbottla: Done..

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
kbottla: Done..

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
kbottla: Done..

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
kbottla: Done..

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
kbottla: Done..

kbottla:
Having tests similar to integration or only integration tests may make it harder to detect issues for any changes in future. I think unit tests (few essential) covering questions above can be useful to quickly detects any issues if new changes are introduced. I have reviewed units tests we had in the past and reduced to few.

Agree mocking may make it hard to read, but it used due to the test to verify no of invocations on a method which returns void.

src/main/resources/config/config.yaml
```diff
@@ -27,10 +27,20 @@ jerseyClientConfig:
   disabledSecureConnection: ${DISABLE_INTERNAL_HTTPS}
 
 rateLimiter:
-  rate: ${RATE_LIMITER_VALUE:-25}  # the rate is for all requests except POST
-  rateForPost: ${RATE_LIMITER_VALUE_POST:-5}
+  rate: ${RATE_LIMITER_VALUE:-75}  # the rate is for all requests except POST. Rate is across all publicapi instances.
```
oswaldquek: this `rate` looks like it should be `numberOfRequests`? the actual rate is calculated by `numberOfRequests` divided by `perMillis`.

src/test/java/uk/gov/pay/api/it/ResourcesFiltersLocalRateLimiterITest.java
```diff
@@ -0,0 +1,203 @@
...
+import static com.jayway.restassured.RestAssured.given;
+import static com.jayway.restassured.http.ContentType.JSON;
+import static io.dropwizard.testing.ConfigOverride.config;
+import static io.dropwizard.testing.ResourceHelpers.resourceFilePath;
+import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
+import static org.hamcrest.core.IsCollectionContaining.hasItem;
+import static org.junit.Assert.assertThat;
+
+/**
+ * PaymentResourceITestBase not extended as RedisDockerRule should not apply for this.
```
oswaldquek: don't need this comment here

src/test/java/uk/gov/pay/api/it/ResourcesFiltersLocalRateLimiterITest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    @Rule
+    public MockServerRule publicAuthMockRule = new MockServerRule(this);
+
+    @Rule
+    public DropwizardAppRule<PublicApiConfig> app = new DropwizardAppRule<>(
+            PublicApi.class
+            , resourceFilePath("config/test-config.yaml")
+            , config("connectorUrl", connectorBaseUrl())
+            , config("publicAuthUrl", publicAuthBaseUrl())
```
oswaldquek: it would be good to add `config("redis.endpoint", "http://bad.url")` to show we're really falling back to the local rate limiting

src/test/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiterTest.java
```diff
@@ -0,0 +1,89 @@
...
+        String key = "Key4";
+        redisRateLimiter = new RedisRateLimiter(2, 2, 1000, jedisPool);
+
+        when(jedis.incr(anyString())).thenReturn(1L, 2L, 1L, 2L);
+
+        redisRateLimiter.checkRateOf(key, POST);
+        redisRateLimiter.checkRateOf(key, POST);
+
+        redisRateLimiter.checkRateOf(key, POST);
+        redisRateLimiter.checkRateOf(key, POST);
```
oswaldquek: i don't understand why this test passes. we're making 4 calls here.

src/test/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiterTest.java
```diff
@@ -0,0 +1,89 @@
...
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class RedisRateLimiterTest {
+
+    @Mock
+    JedisPool jedisPool;
+    @Mock
+    Jedis jedis;
```
oswaldquek: Using mocks is discouraged. Can we use https://github.com/50onRed/mock-jedis which has real implementations of MockJedisPool and Jedis please.
Setting up expectations is prone to error and as commented below, I don't see how rateLimiterSetTo_2CallsPerSecond_shouldAllowMaking3CallsWithinTheAllowedTimeWithSameKey should pass. But when I use the MockJedisPool and MockJedis classes the test fails.

src/test/java/uk/gov/pay/api/filter/ratelimit/RateLimiterTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
+public class RateLimiterTest {
```
oswaldquek: This test can be got rid of if we use a real LocalRateLimiter and RedisRateLimiter in the RateLimiterFilterTest.

oswaldquek:
ok, i can see the value of LocalRateLimiterTest and RedisRateLimiterTest but we should avoid using mocks where possible. added some comments.

src/test/java/uk/gov/pay/api/it/ResourcesFiltersLocalRateLimiterITest.java
kbottla: Agree, added random url

src/test/java/uk/gov/pay/api/filter/ratelimit/RateLimiterTest.java
kbottla: Keeping it as discussed, as it is checking the switch between Local and Redis rate limiters.

src/test/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiterTest.java
kbottla: Removed as it confusing due to mocking Jedis.

PR #245, opened 2018-08-16
PP-4095 Introduce builder pattern for CreatePaymentRequest

alexbishop1:
with @DanailMinchev

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
@@ -8,34 +8,65 @@ import org.apache.commons.lang3.StringUtils;
 @ApiModel(value = "CreatePaymentRequest", description = "The Payment Request Payload")
 public class CreatePaymentRequest {
 
-    public static final String AMOUNT_FIELD_NAME="amount";
-    public static final String RETURN_URL_FIELD_NAME="return_url";
-    public static final String REFERENCE_FIELD_NAME="reference";
-    public static final String DESCRIPTION_FIELD_NAME="description";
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
```diff
@@ -8,34 +8,65 @@
...
+    public static final String REFERENCE_FIELD_NAME = "reference";
+    public static final String DESCRIPTION_FIELD_NAME = "description";
+    public static final String AGREEMENT_ID_FIELD_NAME = "agreement_id";
 
     private int amount;
     private String returnUrl;
     private String reference;
     private String description;
-    private String agreementId = null;
+    private String agreementId;
```
danworth: Can these be final?

src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
```diff
@@ -8,34 +8,65 @@
...
+            return new CreatePaymentRequest(amount, returnUrl, reference, description, agreementId);
+        }
     }
 
-    public CreatePaymentRequest(int amount, String returnUrl, String reference, String description, String agreementId) {
+    public static CreatePaymentRequestBuilder builder() {
+        return new CreatePaymentRequestBuilder();
+    }
+
+    private CreatePaymentRequest(int amount, String returnUrl, String reference, String description, String agreementId) {
```
danworth: This now private constructor could take the builder as a single argument to make it simpler?

DanailMinchev:
As discussed we will address these comments in the next PR, thanks 👍

PR #246, opened 2018-08-16
PP-4095 Add language to CreatePaymentRequest object

alexbishop1:
- Add optional language to `CreatePaymentRequest`
- Deserialise the `"language"` field from the JSON into the object, checking that it’s a string and returning an P0102 “Invalid attribute value” error if it isn’t
- As language is optional, don’t return a P0101 “Missing mandatory attribute” error if the JSON explicitly contains `"language": null`, instead return a P0102 error
- No validation that the field actually contains a supported language yet
- The language in not yet sent to connector when creating the payment

with @DanailMinchev

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
@@ -13,12 +13,14 @@ public class CreatePaymentRequest {
     public static final String REFERENCE_FIELD_NAME = "reference";
     public static final String DESCRIPTION_FIELD_NAME = "description";
     public static final String AGREEMENT_ID_FIELD_NAME = "agreement_id";
+    public static final String LANGUAGE_FIELD_NAME = "language";
 
-    private int amount;
-    private String returnUrl;
-    private String reference;
-    private String description;
-    private String agreementId;
+    private final int amount;
+    private final String returnUrl;
+    private final String reference;
+    private final String description;
+    private final String agreementId;
+    private final String language;
 
     public static class CreatePaymentRequestBuilder {
         private int amount;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
```diff
@@ -13,19 +13,22 @@
...
-    private String returnUrl;
-    private String reference;
-    private String description;
-    private String agreementId;
+    private final int amount;
+    private final String returnUrl;
+    private final String reference;
+    private final String description;
+    private final String agreementId;
+    private final String language;
```
danworth: Isn't there a language enum in common library for this? Would it be better to use that than String?

src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
```diff
@@ -52,21 +55,27 @@ public CreatePaymentRequestBuilder agreementId(String agreementId) {
...
         return new CreatePaymentRequestBuilder();
     }
 
-    private CreatePaymentRequest(int amount, String returnUrl, String reference, String description, String agreementId) {
-        this.amount = amount;
-        this.returnUrl = returnUrl;
-        this.reference = reference;
-        this.description = description;
-        this.agreementId = agreementId;
+    private CreatePaymentRequest(CreatePaymentRequestBuilder createPaymentRequestBuilder) {
```
danworth: Effective Java :)

src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
alexbishop1: That’s coming in the next PR!

There’s two levels of checking of the payload. The first is done as part of the JSON deseralisation: it basically just checks the JSON has fields with the right names and that each field is of the appropriate basic JSON type (string etc.).

Later validation (coming in the next PR) does more semantic checks, such as that the URLs are valid, the amount is within the appropriate range etc. This is where we’ll check the language is actually a supported code and where we’ll start using the enum.

PR #250, opened 2018-08-20
PP-4095 Send language to and return language from connector

alexbishop1:
When processing a create card payment request, send the supplied `"language"` to connector if it passes validation i.e. is `"en"` (English) or `"cy"` (Welsh). If no `"language"` is supplied, do not include a `"language"` field in the request to connector (connector will default to English).

When retrieving a card payment via either the `/v1/payments/{paymentId}` endpoint, the `/v1/payments` search endpoint, or in response to a create payment request, include a `"language" : "en"` or `"language": "cy"` field depending on what connector sends us.

Commits:
1. Make getters in `ValidCreatePaymentRequest` return `Optional`s if appropriate
2. Include `"language"` field when returning a card payment (updates existing create payment Pact test)
3. Send `"language"` field to connector when creating a payment if appropriate (adds new Pact for creating a payment with `"language": "cy"`)

with @DanailMinchev, @danworth

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CardPayment.java
+++ b/src/main/java/uk/gov/pay/api/model/CardPayment.java
@@ -2,15 +2,17 @@ package uk.gov.pay.api.model;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
+import uk.gov.pay.commons.model.SupportedLanguage;
 
 import static uk.gov.pay.api.model.TokenPaymentType.CARD;
 
-
 @JsonInclude(value = JsonInclude.Include.NON_NULL)
 @ApiModel(value = "CardPayment")
-public class CardPayment extends Payment  {
+public class CardPayment extends Payment {
 
     @JsonProperty("refund_summary")
     private final RefundSummary refundSummary;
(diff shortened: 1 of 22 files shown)


src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
```diff
@@ -114,8 +115,8 @@ public void deserialize_shouldDeserializeARequestWithAnAgreementId() throws Exce
         assertThat(paymentRequest.getAmount(), is(27432));
         assertThat(paymentRequest.getReference(), is("Some reference"));
         assertThat(paymentRequest.getDescription(), is("Some description"));
-        assertThat(paymentRequest.getReturnUrl(), is(nullValue()));
-        assertThat(paymentRequest.getAgreementId(), is("abc123"));
+        assertThat(paymentRequest.getReturnUrl(), is(Optional.empty()));
```
georgeracu: Another option might be `assertThat(paymentRequest.getReturnUrl().isPresent(), is(false));`

src/main/java/uk/gov/pay/api/model/CardPayment.java
```diff
@@ -60,6 +66,11 @@ public CardDetails getCardDetails() {
         return cardDetails;
     }
 
+    @ApiModelProperty(example = "en")
```
georgeracu: Maybe adding `dataType` will make it a little more consistent with the rest and will add a little more info regarding the type

src/main/java/uk/gov/pay/api/utils/CustomSupportedLanguageDeserializer.java
```diff
@@ -0,0 +1,24 @@
+package uk.gov.pay.api.utils;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import uk.gov.pay.commons.model.SupportedLanguage;
+
+import java.io.IOException;
+
+public class CustomSupportedLanguageDeserializer extends StdDeserializer<SupportedLanguage> {
```
georgeracu: Something that we should put into the `commons` project

src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
```diff
@@ -67,19 +69,54 @@ public void testCreatePayment() {
                 .build());
 
         PaymentWithAllLinks paymentResponse = createPaymentService.create(account, requestPayload);
+        CardPayment payment = (CardPayment) paymentResponse.getPayment();
```
georgeracu: 👍

src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
alexbishop1: Yep, many ways to skin a cat! 🐈

PR #251, opened 2018-08-21
PP-4073: 429 response code and specify classes to use for swagger.json

kbottla:
PP-4073: limit which classes are exported to swagger docs. 
  -  Configured swagger-maven-plugin to exclude direct debit related endpoints and fields (as we don't want to publish these yet) while generating api docs.
**_with @heathd_** 

PP-4073: Add 429 response code to resource endpoints
   - Added 429 response code to all applicable endpoints.

## 
swagger.json file generated should now 
 - exclude direct debit related api documentation and 
 - contain 429 responses for all endpoints published

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/PaymentState.java
+++ b/src/main/java/uk/gov/pay/api/model/PaymentState.java
@@ -11,7 +11,7 @@ import java.util.Objects;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
-@ApiModel(value="PaymentState", description = "A structure representing the current state of the payment in its lifecycle.")
+@ApiModel(value = "PaymentState", description = "A structure representing the current state of the payment in its lifecycle.")
 public class PaymentState {
     @JsonProperty("status")
     private String status;
@@ -27,10 +27,10 @@ public class PaymentState {
 
     public static PaymentState createPaymentState(JsonNode node) {
         return new PaymentState(
-            node.get("status").asText(),
-            node.get("finished").asBoolean(),
-            node.has("message") ? node.get("message").asText() : null,
-            node.has("code") ? node.get("code").asText() : null
+                node.get("status").asText(),
+                node.get("finished").asBoolean(),
+                node.has("message") ? node.get("message").asText() : null,
+                node.has("code") ? node.get("code").asText() : null
         );
     }
 
@@ -48,22 +48,22 @@ public class PaymentState {
         this.code = code;
     }
 
-    @ApiModelProperty(value = "Current progress of the payment in its lifecycle", required=true, example = "created")
+    @ApiModelProperty(value = "Current progress of the payment in its lifecycle", example = "created")
     public String getStatus() {
         return status;
     }
 
-    @ApiModelProperty(value = "Whether the payment has finished", required = true)
+    @ApiModelProperty(value = "Whether the payment has finished")
     public boolean isFinished() {
         return finished;
     }
 
-    @ApiModelProperty(value = "What went wrong with the Payment if it finished with an error - English message", required = true, example = "User cancelled the payment")
+    @ApiModelProperty(value = "What went wrong with the Payment if it finished with an error - English message", example = "User cancelled the payment")
     public String getMessage() {
         return message;
     }
 
-    @ApiModelProperty(value = "What went wrong with the Payment if it finished with an error - error code", required = true, example = "P010")
+    @ApiModelProperty(value = "What went wrong with the Payment if it finished with an error - error code", example = "P010")
     public String getCode() {
         return code;
     }
--- a/src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.api.model;
 
+import io.swagger.annotations.ApiModelProperty;
 import uk.gov.pay.commons.model.SupportedLanguage;
 
 import java.util.Objects;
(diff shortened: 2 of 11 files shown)


src/main/java/uk/gov/pay/api/model/PaymentState.java
```diff
@@ -93,4 +95,24 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(status, finished, message, code);
     }
+
+    /* No-op setters are required to generate valid swagger.json file. Without the setters, attributes are generated as readonly and mandatory (which is invalid) in swagger file.
+       https://github.com/kongchen/swagger-maven-plugin/issues/458.
+       
+       @JsonProperty removed from variables as Jackson uses setters available to construct object. 
+       Instead constructor has been marked with @JsonCreator and @JsonProperty annotations and getters with @JsonGetter.
```
oswaldquek: what's the benefit of using `@JsonGetter` over `@JsonCreator`? i'm dubious about changing this as it's different from the way we annotation other classes, e.g. uk.gov.pay.api.model.[Payment|PaymentEvent] etc.
also i don't like the no-op setters.

src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
```diff
@@ -50,10 +51,12 @@ public String getDescription() {
         return Optional.ofNullable(returnUrl);
     }
 
+    @ApiModelProperty(name = "agreementId", access = "agreementId")
     public Optional<String> getAgreementId() {
         return Optional.ofNullable(agreementId);
     }
 
+    @ApiModelProperty(name = "language", access = "language")
```
oswaldquek: what has this line got to do with the goal of this story, i.e. "PP-7073 429 response code"?

src/test/java/uk/gov/pay/api/resources/error/ApiErrorResponseTest.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.pay.api.resources.error;
+
+import org.junit.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+public class ApiErrorResponseTest {
```
oswaldquek: This test is fine but i think it should be tested on an integration test as the user need is that the calling client will be shown the error message?

src/main/java/uk/gov/pay/api/model/PaymentState.java
kbottla: We are using both @JsonGetter and @JsonCreator for this change. It is because of the issue with swagger plugin. Swagger generates invalid swagger.json if no-op setters are not available (generates fields as read only which is not the case). 

And if no-op setters are introduced, Jackson doesn't like the way we usually use @JsonProperty.

src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
kbottla: It is part of removing direct debit related changes. We need to define an ApiModelProperty and use it in pom.xml config for swagger-maven-plugin to ignore certain fields.

src/test/java/uk/gov/pay/api/resources/error/ApiErrorResponseTest.java
kbottla: we may already have integration tests for 429 errors : https://github.com/alphagov/pay-publicapi/blob/pp_4073_429_response_for_endpoints/src/test/java/uk/gov/pay/api/it/ResourcesFilterRateLimiterITest.java#L36

src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
oswaldquek: ok 👍

src/main/java/uk/gov/pay/api/resources/error/ApiErrorResponse.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public static ApiErrorResponse anApiErrorResponse(Code code) {
+        return new ApiErrorResponse(code);
+    }
+
+    public static ApiErrorResponse anApiErrorResponse(Code code, Object... parameters) {
+        return new ApiErrorResponse(code, parameters);
+    }
+
+    private ApiErrorResponse(Code code, Object... parameters) {
```
oswaldquek: this method is never used

src/test/java/uk/gov/pay/api/resources/error/ApiErrorResponseTest.java
oswaldquek: ok. then i'm not so sure why we need this unit test as it feels like an extra unit test for the sake of it. this is just my opinion though. it'd be good to get a consensus from the rest of the dev team as to whether we need double testing at integration and unit level.

src/main/java/uk/gov/pay/api/resources/error/ApiErrorResponse.java
kbottla: This is used in RateLimiterFilter. 

https://github.com/alphagov/pay-publicapi/blob/pp_4073_429_response_for_endpoints/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java#L70

src/main/java/uk/gov/pay/api/model/PaymentState.java
kbottla: Removed 'readOnly' for attributes. It is now inline with the rest of our code.

src/main/java/uk/gov/pay/api/resources/error/ApiErrorResponse.java
oswaldquek: oh oops. looks like i was viewing an older commit locally.

src/test/java/uk/gov/pay/api/resources/error/ApiErrorResponseTest.java
kbottla: This is added inline with https://github.com/alphagov/pay-publicapi/blob/pp_4073_429_response_for_endpoints/src/test/java/uk/gov/pay/api/model/PaymentErrorTest.java we had in place.

In this case I see your point that unit test is not adding much value as toString() is never used. So removed this test.

PR #255, opened 2018-09-04
PP-4106 Add delayedCapture field to create payment request

DanailMinchev:
## WHAT

- Added `delayedCapture` property to create payment request classes
- Modified `RequestJsonParser` and related tests
- Added missing `@Test` annotations and fixed the related tests

with @SandorArpa

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
@@ -17,6 +17,7 @@ public class CreatePaymentRequest {
     public static final String DESCRIPTION_FIELD_NAME = "description";
     public static final String AGREEMENT_ID_FIELD_NAME = "agreement_id";
     public static final String LANGUAGE_FIELD_NAME = "language";
+    public static final String DELAYED_CAPTURE_FIELD_NAME = "delayed_capture";
 
     private final int amount;
     private final String returnUrl;
@@ -24,6 +25,7 @@ public class CreatePaymentRequest {
     private final String description;
     private final String agreementId;
     private final String language;
+    private final Boolean delayedCapture;
 
     public static class CreatePaymentRequestBuilder {
         private int amount;
@@ -32,6 +34,7 @@ public class CreatePaymentRequest {
         private String description;
         private String agreementId;
         private String language;
+        private Boolean delayedCapture;
 
         public CreatePaymentRequestBuilder amount(int amount) {
             this.amount = amount;
@@ -63,6 +66,11 @@ public class CreatePaymentRequest {
             return this;
         }
 
+        public CreatePaymentRequestBuilder delayedCapture(Boolean delayedCapture) {
+            this.delayedCapture = delayedCapture;
+            return this;
+        }
+
         public CreatePaymentRequest build() {
             return new CreatePaymentRequest(this);
         }
@@ -79,6 +87,7 @@ public class CreatePaymentRequest {
         this.description = createPaymentRequestBuilder.description;
         this.agreementId = createPaymentRequestBuilder.agreementId;
         this.language = createPaymentRequestBuilder.language;
+        this.delayedCapture = createPaymentRequestBuilder.delayedCapture;
     }
 
     @ApiModelProperty(value = "amount in pence", required = true, allowableValues = "range[1, 10000000]", example = "12000")
@@ -114,6 +123,12 @@ public class CreatePaymentRequest {
         return language;
     }
 
+    @ApiModelProperty(value = "delayed capture flag", required = false, example = "false")
+    @JsonProperty(DELAYED_CAPTURE_FIELD_NAME)
+    public Boolean getDelayedCapture() {
+        return delayedCapture;
+    }
+
     public boolean hasReturnUrl() {
         return StringUtils.isNotBlank(returnUrl);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
```diff
@@ -126,6 +141,10 @@ public boolean hasLanguage() {
         return StringUtils.isNotBlank(language);
     }
 
+    public boolean hasDelayedCapture() {
+        return delayedCapture != null;
+    }
```
alexbishop1: This method doesn’t appear to be used

src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
```diff
@@ -62,6 +62,7 @@ public void deserialize_shouldDeserializeARequestWithReturnUrlSuccessfully() thr
         assertThat(paymentRequest.getReturnUrl().get(), is("https://somewhere.gov.uk/rainbow/1"));
     }
 
+    @Test
```
alexbishop1: 🤦‍♂️

src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
```diff
@@ -96,12 +98,52 @@ public void deserialize_shouldDeserializeARequestWithWelshLanguageSuccessfully()
...
+                "  \"delayed_capture\": false\n" +
+                "}";
+
+        ValidCreatePaymentRequest paymentRequest = deserializer.deserialize(jsonFactory.createParser(validJson), ctx);
+
+        assertThat(paymentRequest.getAmount(), is(27432));
+        assertThat(paymentRequest.getReference(), is("Some reference"));
+        assertThat(paymentRequest.getDescription(), is("Some description"));
+        assertThat(paymentRequest.getReturnUrl().get(), is("https://somewhere.gov.uk/rainbow/1"));
+        assertThat(paymentRequest.getDelayedCapture().get(), is(Boolean.FALSE));
```
alexbishop1: I wonder if it might be worth adding something to `deserialize_shouldDeserializeARequestWithReturnUrlSuccessfully` (the test that doesn’t include any optional properties) to assert that `paymentRequest.getDelayedCapture()` and `paymentRequest.getLanguage()` both return `Optional.empty()` i.e. that we’re not doing any defaulting at this stage.

src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
```diff
@@ -209,7 +251,7 @@ public void deserialize_shouldThrowValidationException_whenAmountIsMoreThanMaxim
     }
 
     @Test
-    public void deserialize_shouldThrowValidationException_whenReturnUrlIsNotAnStringValue() throws Exception {
+    public void deserialize_shouldThrowValidationException_whenReturnUrlIsNotAStringValue() throws Exception {
```
alexbishop1: 👍

src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
```diff
@@ -509,6 +551,86 @@ public void deserialize_shouldThrowValidationException_whenLanguageIsEmptyString
...
+                "  \"description\": \"Some description\",\n" +
+                "  \"return_url\": \"https://somewhere.gov.uk/rainbow/1\",\n" +
+                "  \"delayed_capture\": 0\n" +
+                "}";
+
+        expectedException.expect(aBadRequestExceptionWithError("P0102", "Invalid attribute value: delayed_capture. Must be true or false"));
+
+        deserializer.deserialize(jsonFactory.createParser(json), ctx);
+    }
+
```
alexbishop1: These are exhaustive but I’m not sure we need all of them. Probably just checking `null`, a single not-a-boolean (e.g. a number or a string) and maybe a string literal of `"true"` should be enough.

src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
```diff
@@ -509,6 +553,54 @@ public void deserialize_shouldThrowValidationException_whenLanguageIsEmptyString
         deserializer.deserialize(jsonFactory.createParser(json), ctx);
     }
 
+    @Test
+    public void deserialize_shouldThrowValidationException_whenDelayedCaptureIsAString() throws Exception {
```
alexbishop1: Minor, but I think `IsNotABoolean` (rather than `IsAString`) might capture the intent of this test better

src/test/java/uk/gov/pay/api/json/RequestJsonParserTest.java
```diff
@@ -37,40 +37,42 @@ public void parsePaymentRequest_withReturnUrl_shouldParseSuccessfully() throws E
...
         // language=JSON
         String payload = "{\n" +
-                "  \"amount\": -1000,\n" +
+                "  \"amount\": 1000,\n" +
                 "  \"reference\": \"Some reference\",\n" +
                 "  \"description\": \"Some description\",\n" +
                 "  \"return_url\": \"https://somewhere.gov.uk/rainbow/1\",\n" +
-                "  \"language\": \"en\"\n" +
+                "  \"language\": \"en\",\n" +
+                "  \"delayed_capture\": false\n" +
```
alexbishop1: I think I’d prefer if this were `true` just to make us more confident that we’re not somehow relying on the default value of a `boolean` field being `false`

PR #256, opened 2018-09-06
PP-4106 Delayed capture connector integration

DanailMinchev:
## WHAT

- Add `delayedCapture` property to the card model
- Make the call to connector include the `delayed_capture` field if appropriate
- Ensure the Public API endpoints for creating payments and retrieving payments return the `delayed_capture` field

The next PR will have Pact tests

with @SandorArpa

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CardPayment.java
+++ b/src/main/java/uk/gov/pay/api/model/CardPayment.java
@@ -24,19 +24,24 @@ public class CardPayment extends Payment {
     private final CardDetails cardDetails;
 
     @JsonSerialize(using = ToStringSerializer.class)
-    @ApiModelProperty(name = "language", access = "language") // named to exclude from swagger.json for now
+    @ApiModelProperty(name = "language", access = "language")
     private final SupportedLanguage language;
+
+    @JsonProperty("delayed_capture")
+    @ApiModelProperty(name = "delayed_capture", access = "delayed_capture")
+    private final boolean delayedCapture;
 
     public CardPayment(String chargeId, long amount, PaymentState state, String returnUrl, String description,
                        String reference, String email, String paymentProvider, String createdDate,
                        RefundSummary refundSummary, SettlementSummary settlementSummary, CardDetails cardDetails,
-                       SupportedLanguage language) {
+                       SupportedLanguage language, boolean delayedCapture) {
         super(chargeId, amount, state, returnUrl, description, reference, email, paymentProvider, createdDate);
         this.refundSummary = refundSummary;
         this.settlementSummary = settlementSummary;
         this.cardDetails = cardDetails;
         this.paymentType = CARD.getFriendlyName();
         this.language = language;
+        this.delayedCapture = delayedCapture;
     }
 
     /**
@@ -72,6 +77,11 @@ public class CardPayment extends Payment {
         return language;
     }
 
+    @ApiModelProperty(value = "delayed capture flag", example = "false")
+    public boolean getDelayedCapture() {
+        return delayedCapture;
+    }
+
     @Override
     public String toString() {
         // Some services put PII in the description, so don’t include it in the stringification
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/pay/api/model/CardPayment.java
```diff
@@ -84,6 +93,7 @@ public String toString() {
                 ", returnUrl='" + returnUrl + '\'' +
                 ", reference='" + reference + '\'' +
                 ", language='" + language.toString() + '\'' +
+                ", delayedCapture=" + (delayedCapture != null ? delayedCapture.toString() : "") +
```
alexbishop1: Shouldn’t need the `null` check here: now that PP-4105 is done, connector should always return `delayed_capture`

src/test/java/uk/gov/pay/api/it/CreatePaymentITest.java
```diff
@@ -45,11 +45,10 @@
 
     @Test
     public void createCardPayment() {
-
         publicAuthMock.mapBearerTokenToAccountId(API_KEY, GATEWAY_ACCOUNT_ID, CARD);
 
         connectorMock.respondOk_whenCreateCharge(AMOUNT, GATEWAY_ACCOUNT_ID, CHARGE_ID, CHARGE_TOKEN_ID,
-                CREATED, RETURN_URL, DESCRIPTION, REFERENCE, null, PAYMENT_PROVIDER, CREATED_DATE, SupportedLanguage.ENGLISH, REFUND_SUMMARY,
+                CREATED, RETURN_URL, DESCRIPTION, REFERENCE, null, PAYMENT_PROVIDER, CREATED_DATE, SupportedLanguage.ENGLISH, Boolean.FALSE, REFUND_SUMMARY,
```
alexbishop1: (This applies in a lot of places)

I think it’s preferable to use `true` for test booleans to avoid the default for `boolean` fields as being `false`: this checks that the value is actually being passed all the way through

src/test/java/uk/gov/pay/api/it/validation/PaymentsRefundsResourceAmountValidationITest.java
```diff
@@ -272,7 +272,7 @@ public void createPaymentRefund_responseWith400_whenConnectorResponseIsErrorDueT
         String externalChargeId = "charge_12345";
 
         connectorMock.respondWithChargeFound(amount, GATEWAY_ACCOUNT_ID, externalChargeId, null, null, null, null,
-                null, null, null,  SupportedLanguage.ENGLISH, null,
+                null, null, null,  SupportedLanguage.ENGLISH, Boolean.FALSE,null,
```
alexbishop1: No space between the commas! 😱

src/test/java/uk/gov/pay/api/utils/mocks/ConnectorMockClient.java
```diff
@@ -77,6 +77,7 @@ private String buildChargeResponse(long amount, String chargeId, PaymentState st
                 .add("card_brand", CARD_BRAND_LABEL)
                 .add("created_date", createdDate)
                 .add("language", language.toString())
+                .add("delayed_capture", Boolean.toString(delayedCapture))
```
alexbishop1: If it’s a `Boolean` object, you can just `toString()` the object directly: `delayedCapture.toString()`.

This works if it’s a primitive though.

src/test/java/uk/gov/pay/api/it/validation/PaymentsRefundsResourceAmountValidationITest.java
DanailMinchev: 🙀

DanailMinchev: I missed adding `// language=JSON` everywhere as well!

src/test/java/uk/gov/pay/api/utils/mocks/ConnectorMockClient.java
alexbishop1: But, of course, as demonstrated in your latest code, we actually want the value itself, not its stringification!

PR #258, opened 2018-09-10
PP-4106 add get payment with delayed capture pact integration

SandorArpa:
## WHAT:
Add `get payment with delayed capture true` pact test.

## HOW:
- Refactor PaymentsResource and add GetPaymentService
- Add Pact tests
- Create a json file for 'get a payment with delayed capture true'

with @DanailMinchev

### DIFF
--- a/src/test/java/uk/gov/pay/api/service/GetPaymentServiceTest.java
+++ b/src/test/java/uk/gov/pay/api/service/GetPaymentServiceTest.java
@@ -1,0 +1,90 @@
+package uk.gov.pay.api.service;
+
+import au.com.dius.pact.consumer.PactVerification;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/api/service/GetPaymentServiceTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Rule
+    public PactProviderRule connectorRule = new PactProviderRule("connector", this);
+
+    @Mock
+    private PublicApiConfig mockConfiguration;
+
+    @Before
+    public void setup() {
+        // We will actually send real requests here, which will be intercepted by pact
```
oswaldquek: I don't think we need this comment here as it's clear from line 48 we're returning a non-mock endpoint created by the PactProviderRule.

DanailMinchev:
Thanks for the review @oswaldquek

As discussed with @oswaldquek @alexbishop1 and @SandorArpa we still need `GetPaymentITest.java` (mocked integration test) as it tests all of the properties and we have `GetPaymentServiceTest` (Pact integration test) as it tests only the integration which is needed by the client.

PR #263, opened 2018-09-12
Pp 4113 basic get refunds endpoint

cobainc0:
## WHAT
Adds refunds resource to return all refunds for a service provider,
which can be queried by page number and display size. 
Also adds new Pact interactions. 

## HOW 
_Steps to test or reproduce:_
Pact tests for refunds SearchRefundsServiceTest


src/test/resources/pacts/publicapi-connector-search-refunds-with-page-and-display-when-no-refunds-exist.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "provider": {
+    "name": "connector"
+  },
+  "interactions": [
+    {
+      "description": "Return no refunds when no refunds exist",
+      "providerStates": [
+        {
+          "name": "Account exists and has no refunds"
```
oswaldquek: you probably want to add a "params" block here with `"gateway_account_id": 888`. see https://github.com/alphagov/pay-publicapi/blob/75366929a6b6868b27ac349cc3408e03ae7e67a8/src/test/resources/pacts/publicapi-connector-get-payment-with-delayed-capture-true.json#L14 as an example.

src/test/resources/pacts/publicapi-connector-search-refunds-with-page-and-display.json
```diff
@@ -0,0 +1,149 @@
...
+  },
+  "provider": {
+    "name": "connector"
+  },
+  "interactions": [
+    {
+      "description": "Return all refunds when refunds exist",
+      "providerStates": [
+        {
+          "name": "Account exists and has 2 refunds"
```
oswaldquek: you probably want to add a "params" block here with `"gateway_account_id": 777`. see https://github.com/alphagov/pay-publicapi/blob/75366929a6b6868b27ac349cc3408e03ae7e67a8/src/test/resources/pacts/publicapi-connector-get-payment-with-delayed-capture-true.json#L14 as an example.

src/main/java/uk/gov/pay/api/model/search/card/SearchRefunds.java
```diff
@@ -0,0 +1,131 @@
...
+
+    protected HalRepresentation.HalRepresentationBuilder decoratePagination(Long accountId,
+                                                                            HalRepresentation.HalRepresentationBuilder halRepresentationBuilder,
+                                                                            SearchRefundsResponse pagination) {
+        try {
+            halRepresentationBuilder
+                    .addProperty("count", pagination.getCount())
+                    .addProperty("total", pagination.getTotal())
+                    .addProperty("page", pagination.getPage());
+            addLink(halRepresentationBuilder, "self", transformIntoPublicUri(
```
simad: i see there's already logic for pagination for search payments, if it's the same we should definitely move that up the hierarchy and make both payments and refund use it - we don't want to have pagination logic in two places.

The difference I can see is that you removed first, last, prev and next pages from here - is this by design? What happens if we have multiple pages?

src/main/java/uk/gov/pay/api/model/search/card/SearchRefunds.java
```diff
@@ -0,0 +1,131 @@
...
+        return halRepresentationBuilder;
+    }
+
+    private void addLink(HalRepresentation.HalRepresentationBuilder halRepresentationBuilder, String name, URI uri) {
+        if (uri != null) {
+            halRepresentationBuilder.addLink(name, uri);
+        }
+    }
+
+    private URI transformIntoPublicUri(String baseUrl, Link link, Long accountId) throws URISyntaxException {
```
simad: same as above, this duplicates a method in https://github.com/alphagov/pay-publicapi/blob/f16da88549f0bd4a8f2063a06876f8e78cec154f/src/main/java/uk/gov/pay/api/model/search/SearchPaymentsBase.java

We might want payents and refund to inherit from a common base class, otherwise we are going to duplicate lots of 💩

src/main/java/uk/gov/pay/api/model/search/card/SearchRefundsResults.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.pay.api.model.search.card;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
```
simad: so many intermediate objects :( not your fault, but it's growing a bit out of control. We might want to address this, but not as part of this story, I guess

src/main/java/uk/gov/pay/api/service/SearchRefundsService.java
```diff
@@ -0,0 +1,84 @@
...
+                                ConnectorUriGenerator uriGenerator,
+                                ObjectMapper objectMapper) {
+
+        this.client = client;
+        this.configuration = configuration;
+        this.uriGenerator = uriGenerator;
+        this.objectMapper = objectMapper;
+    }
+
+    public Response getAllRefunds(Account account, String page, String displaySize) {
```
simad: we might want to have a class (maybe an inner one, maybe not) for RefundSearchParams. Right now we only have page and display size, but this will grow and we don't want this method to become a monster with 100 parameters. This should also avoid you having to stick stuff in a map, which is always a bit ...sketchy

src/main/java/uk/gov/pay/api/service/SearchRefundsService.java
```diff
@@ -0,0 +1,84 @@
...
+        return refundsService.getSearchResponse(account, queryParams);
+    }
+
+    public void validateRefundSearchParameters(String page, String displaySize) {
+        List<String> validationErrors = new LinkedList<>();
+        try {
+            validatePageIfNotNull(page, validationErrors);
+            validateDisplaySizeIfNotNull(displaySize, validationErrors);
+        } catch (Exception e) {
+            throw new ValidationException(aPaymentError(SEARCH_PAYMENTS_VALIDATION_ERROR, join(validationErrors, ", "), e.getMessage()));
```
simad: This error is generic enough I would rename it SEARCH_VALIDATION_ERROR, as it is not specific to payments anymore

src/main/java/uk/gov/pay/api/service/SearchRefundsService.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+        if (!validationErrors.isEmpty()) {
+            throw new ValidationException(aPaymentError(SEARCH_PAYMENTS_VALIDATION_ERROR, join(validationErrors, ", ")));
+        }
+    }
+
+    private void validatePageIfNotNull(String pageNumber, List<String> validationErrors) {
```
simad: Same as all considerations above, if we have common stuff in validators they should be shared. Is there a usecase for having different max display size (for example) for search or refunds?

src/main/java/uk/gov/pay/api/model/search/card/SearchRefunds.java
cobainc0: I think we will add this soon as it would mean changing connector and the pact files.

src/main/java/uk/gov/pay/api/model/search/SearchBase.java
```diff
@@ -0,0 +1,65 @@
+package uk.gov.pay.api.model.search;
```
simad: 🏅 nice work with this class

src/main/java/uk/gov/pay/api/service/ConnectorUriGenerator.java
```diff
@@ -35,6 +35,11 @@ public String chargesURI(Account account, String agreementId) {
     public String chargesURIWithParams(Account account, Map<String, String> queryParams) {
         return buildConnectorUri(account, format("/v1/api/accounts/%s/charges", account.getAccountId()), queryParams);
     }
+    
+    public String refundsURIWithParams(Account account, Map<String, String> queryParams) {
+        String path = format("/v1/accounts/%s/refunds", account.getAccountId());
```
simad: This does not match the endpoint in connector :(

src/main/java/uk/gov/pay/api/service/RefundsParams.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private String page;
+    private String displaySize;
+
+    public RefundsParams(String page, String displaySize) {
+        this.page = page;
+        this.displaySize = displaySize;
+    }
+
+    public void setPage(String page) {
```
simad: I would personally not put setters in unless you plan to use them. It's good to have "immutable" classes, especially considering that this should be instantiated after the parameters are parsed and never changed

src/main/java/uk/gov/pay/api/model/search/card/SearchRefunds.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private HalRepresentation.HalRepresentationBuilder decoratePagination(Long accountId,
+                                                                          HalRepresentation.HalRepresentationBuilder halRepresentationBuilder,
+                                                                          ISearchPagination pagination) {
+        HalRepresentation.HalRepresentationBuilder builder = addPaginationProperties(halRepresentationBuilder, pagination);
+        SearchRefundsNavigationLinks links = (SearchRefundsNavigationLinks) pagination.getLinks();
+        String path = getRefundsPath(accountId);
+
+        try {
+            addLink(builder, "self", transformIntoPublicUri(baseUrl, links.getSelfPage(), path));
```
simad: I think `path` should just be `REFUNDS_PATH`, which is  "/v1/refunds". We don't need `/accounts/accountid` in publicapi as the account depends on the bearer token, you can't retrieve resources for other accounts

src/main/java/uk/gov/pay/api/validation/SearchRefundsValidator.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+
+import static org.apache.commons.lang3.StringUtils.join;
+import static uk.gov.pay.api.model.RefundError.Code.SEARCH_REFUNDS_VALIDATION_ERROR;
+import static uk.gov.pay.api.model.RefundError.aRefundError;
+
+
+public class SearchRefundsValidator extends SearchValidator {
+
+    public static void validateSearchParameters(String pageNumber,
```
simad: One of those two methods should be removed, up to you which one to keep :)

src/main/java/uk/gov/pay/api/validation/SearchValidator.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.List;
+
+import static org.eclipse.jetty.util.StringUtil.isNotBlank;
+
+
+public abstract class SearchValidator {
+
+    protected static void validatePageIfNotNull(String pageNumber, List<String> validationErrors) {
+        if (isNotBlank(pageNumber) && (!StringUtils.isNumeric(pageNumber) || Integer.valueOf(pageNumber) < 1)) {
```
simad: There is a NumericValidator in PublicApi for this. If you ask why we are not using isNumeric, help yourself https://github.com/alphagov/pay-publicapi/pull/267#discussion_r220148191

src/main/java/uk/gov/pay/api/validation/SearchValidator.java
```diff
@@ -0,0 +1,23 @@
...
+            validationErrors.add("page");
+        }
+    }
+
+    protected static void validateDisplaySizeIfNotNull(String displaySize, List<String> validationErrors) {
+        if (isNotBlank(displaySize) && (!StringUtils.isNumeric(displaySize) || Integer.valueOf(displaySize) < 1 || Integer.valueOf(displaySize) > 500)) {
```
simad: same,see isNumeric point as above

src/main/java/uk/gov/pay/api/service/ConnectorUriGenerator.java
cobainc0: Updated! Thanks

src/main/java/uk/gov/pay/api/service/RefundsParams.java
cobainc0: ✪🌟⭒💫✶✨✴︎

simad: was this the response you were talking about? it's just a list of emojis 🤔

src/main/java/uk/gov/pay/api/model/search/ISearchPagination.java
```diff
@@ -0,0 +1,12 @@
...
+
+
+public interface ISearchPagination<T> {
+    int getCount();
+
+    int getTotal();
+
+    int getPage();
+
+    T getLinks();
```
simad: We now have `SearchNavigationLinks` as the type of T everywhere, so this interface does not need to be generic anymore. 
Also, we don't have a convention of using `I` to name interfaces, so you can rename this as `SearchPagination`

PR #266, opened 2018-09-20
PP-4144 Return last 4 digits and first 6 digits of card number in get payment

simad:
## WHAT
We are about to add support for storing the first 6 digits of a card number in connector. This PR makes PublicApi return its value if present, or `null` if not.

### DIFF
--- a/swagger/swagger.json
+++ b/swagger/swagger.json
@@ -520,6 +520,11 @@
           "example" : "1234",
           "readOnly" : true
         },
+        "first_digits_card_number" : {
+          "type" : "string",
+          "example" : "123456",
+          "readOnly" : true
+        },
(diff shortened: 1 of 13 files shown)


swagger/swagger.json
```diff
@@ -543,77 +548,37 @@
...
-        "_links" : {
-          "readOnly" : true,
-          "$ref" : "#/definitions/PaymentLinksForSearch"
-        },
-        "card_brand" : {
-          "type" : "string",
-          "example" : "Visa",
-          "description" : "Card Brand",
-          "readOnly" : true
+      "allOf" : [ {
```
DanailMinchev: I guess these changes were not synched to `swagger.json` that is why they are not related to the PR

PR #272, opened 2018-10-18
PP-4293 Add filtering by from and to date in the refunds search endpoints

simad:
## WHAT 
- Added two query params to the `/v1/refunds` endpoint, `from_date` and `to_date`, to filter refunds by date range similarly to how we do it for payments
- The api docs have _not_ been updated, will be updated as a separate PR


src/main/java/uk/gov/pay/api/model/search/ISearchPagination.java
```diff
@@ -1,12 +0,0 @@
-package uk.gov.pay.api.model.search;
-
-
-public interface ISearchPagination<T> {
```
DanailMinchev: Looks like in other Java repos we do not use `I` for interface prefixes:

- https://github.com/alphagov/pay-adminusers/search?q=public+interface&unscoped_q=public+interface
- https://github.com/alphagov/pay-direct-debit-connector/search?p=1&q=public+interface&unscoped_q=public+interface
- https://github.com/alphagov/pay-connector/search?q=public+interface&unscoped_q=public+interface

So, I guess we want to rename that and not use this prefix as you did.

You can check with other team members.

simad: I thought I renamed that

simad: Yes one of the changes is renaming the interface

src/main/java/uk/gov/pay/api/model/search/directdebit/DDSearchResponse.java
```diff
@@ -4,13 +4,13 @@
...
 import uk.gov.pay.api.model.links.SearchNavigationLinks;
-import uk.gov.pay.api.model.search.ISearchPagination;
+import uk.gov.pay.api.model.search.SearchPagination;
 
 import java.util.List;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class DDSearchResponse implements ISearchPagination<SearchNavigationLinks> {
+public class DDSearchResponse implements SearchPagination {
```
DanailMinchev: Would you like to rename `DD` to `DirectDebit`? May be check other team members

simad: It seems we have mostly DD for direct debit related things in publicapi, except `SearchDirectDebitPayments` 😬

src/main/java/uk/gov/pay/api/service/SearchRefundsService.java
```diff
@@ -17,12 +16,15 @@
 import java.util.Optional;
 
 import static uk.gov.pay.api.model.RefundError.Code.SEARCH_REFUNDS_DIRECT_DEBIT_ERROR;
+import static uk.gov.pay.api.validation.RefundSearchValidator.*;
```
DanailMinchev: We don't use wildcards usually

src/main/java/uk/gov/pay/api/validation/SearchValidator.java
```diff
@@ -5,17 +5,29 @@
 import static org.eclipse.jetty.util.StringUtil.isNotBlank;
 
 
-public abstract class SearchValidator {
```
DanailMinchev: Not sure why `abstract` was added, is it because we want to make sure that we can't initialise this class directly and we need specific implementations? On the other hand all the methods are static there so we will not use it as an object anyway

src/main/java/uk/gov/pay/api/model/search/ISearchPagination.java
DanailMinchev: Yes, I wanted to clarify why you renamed it:
it's because on other projects we do not use `I` prefixes for interfaces and if you want to clarify why we introduced this naming with the person who did it.
Sorry it was not very clear

simad: Yes, it was introduced by mistake and I am renaming it to a more consistent pattern 👍

src/main/java/uk/gov/pay/api/model/search/directdebit/DDSearchResponse.java
simad: I will rename everything to DirectDebit, ha

src/main/java/uk/gov/pay/api/service/SearchRefundsService.java
simad: sorry, I've setup intellij to not do this but it keeps collapsing imports 😢

src/main/java/uk/gov/pay/api/validation/SearchValidator.java
simad: I think it was abstract because originally the refund class was extending this, but being it a collection of static methods there was no need to do so. It is not abstract anymore

PR #274, opened 2018-10-19
PP-4128 Make CardPayment use Optional

SandorArpa:
## WHAT YOU DID
- refactor CardPayment's and Payment's getters to return Optional where appropriate. This is to avoid using null.


src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
```diff
@@ -75,9 +76,9 @@ public void testCreatePayment() {
         assertThat(payment.getAmount(), is(100L));
         assertThat(payment.getReference(), is("a reference"));
         assertThat(payment.getDescription(), is("a description"));
-        assertThat(payment.getEmail(), is(nullValue()));
+        assertThat(payment.getEmail(), is(Optional.empty()));
```
DanailMinchev: Could you remove unused `org.hamcrest.core.IsNull.nullValue` import as well, thanks

PR #278, opened 2018-10-31
PP-4113 Refund status should be external in api docs

simad:
## WHAT YOU DID
Just updating the docs after fixing a bug in connector.


jonheslop:
🎖 nice one

PR #280, opened 2018-11-01
PP-4297 Add capture url to `/v1/payments/{payment_external_id}` response

SandorArpa:
## WHAT YOU DID
Return capture url in the response object when received from connector
Add/refactor process to handle capture uri

## How to test
When a charge is on AWAITING_CAPTURE_REQUEST state in connector, calling the endpoint `/v1/payments/{payment_external_id}` should include 
```
"_links":
   "capture": {
            "href": "xxxx/v1/payments/xxxxx/capture",
            "method": "POST"
        }
```
in the response payload. 
Any other states should include
```
"_links":
   "capture": null
```

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
+++ b/src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
@@ -42,7 +42,7 @@ public class PaymentWithAllLinks {
                                String reference, String email, String paymentProvider, String createdDate, SupportedLanguage language,
                                boolean delayedCapture, RefundSummary refundSummary, SettlementSummary settlementSummary, CardDetails cardDetails,
                                List<PaymentConnectorResponseLink> paymentConnectorResponseLinks, URI selfLink, URI paymentEventsUri, URI paymentCancelUri,
-                               URI paymentRefundsUri, Long corporateCardSurcharge, Long totalAmount) {
+                               URI paymentRefundsUri, URI captureUri, Long corporateCardSurcharge, Long totalAmount) {
         this.payment = new CardPayment(chargeId, amount, state, returnUrl, description, reference, email, paymentProvider, createdDate,
                 refundSummary, settlementSummary, cardDetails, language, delayedCapture, corporateCardSurcharge, totalAmount);
         this.links.addSelf(selfLink.toString());
@@ -52,6 +52,10 @@ public class PaymentWithAllLinks {
 
         if (!state.isFinished()) {
             this.links.addCancel(paymentCancelUri.toString());
+        }
+        
+        if (paymentConnectorResponseLinks.stream().anyMatch(link -> "capture".equals(link.getRel()))) {
+            this.links.addCapture(captureUri.toString());
         }
     }
 
@@ -84,7 +88,8 @@ public class PaymentWithAllLinks {
                                               URI selfLink,
                                               URI paymentEventsUri,
                                               URI paymentCancelUri,
-                                              URI paymentRefundsUri) {
+                                              URI paymentRefundsUri,
+                                              URI paymentsCaptureUri) {
         return new PaymentWithAllLinks(
                 paymentConnector.getChargeId(),
                 paymentConnector.getAmount(),
@@ -105,6 +110,7 @@ public class PaymentWithAllLinks {
                 paymentEventsUri,
                 paymentCancelUri,
                 paymentRefundsUri,
+                paymentsCaptureUri,
                 paymentConnector.getCorporateCardSurcharge(), 
                 paymentConnector.getTotalAmount()
         );
@@ -116,12 +122,13 @@ public class PaymentWithAllLinks {
             URI selfLink,
             URI paymentEventsUri,
             URI paymentCancelUri,
-            URI paymentRefundsUri) {
+            URI paymentRefundsUri,
+            URI paymentsCaptureUri) {
         switch (paymentType) {
             case DIRECT_DEBIT:
                 return PaymentWithAllLinks.valueOf(paymentConnector, selfLink);
             default:
-                return PaymentWithAllLinks.valueOf(paymentConnector, selfLink, paymentEventsUri, paymentCancelUri, paymentRefundsUri);
+                return PaymentWithAllLinks.valueOf(paymentConnector, selfLink, paymentEventsUri, paymentCancelUri, paymentRefundsUri, paymentsCaptureUri);
         }
     }
 
--- a/src/test/java/uk/gov/pay/api/it/GetPaymentITest.java
+++ b/src/test/java/uk/gov/pay/api/it/GetPaymentITest.java
@@ -40,6 +40,7 @@ public class GetPaymentITest extends PaymentResourceITestBase {
     private static final String CHARGE_TOKEN_ID = "token_1234567asdf";
     private static final PaymentState CREATED = new PaymentState("created", false, null, null);
     private static final PaymentState CAPTURED = new PaymentState("captured", false, null, null);
+    public static final PaymentState AWAITING_CAPTURE_REQUEST = new PaymentState("submitted", false, null, null);
     private static final RefundSummary REFUND_SUMMARY = new RefundSummary("pending", 100L, 50L);
     private static final String PAYMENT_PROVIDER = "Sandbox";
     private static final String CARD_BRAND_LABEL = "Mastercard";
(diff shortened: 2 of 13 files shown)


src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
```diff
@@ -42,7 +42,7 @@ public PaymentWithAllLinks(String chargeId, long amount, PaymentState state, Str
                                String reference, String email, String paymentProvider, String createdDate, SupportedLanguage language,
                                boolean delayedCapture, RefundSummary refundSummary, SettlementSummary settlementSummary, CardDetails cardDetails,
                                List<PaymentConnectorResponseLink> paymentConnectorResponseLinks, URI selfLink, URI paymentEventsUri, URI paymentCancelUri,
-                               URI paymentRefundsUri, Long corporateCardSurcharge, Long totalAmount) {
+                               URI paymentRefundsUri, URI captureUri, Long corporateCardSurcharge, Long totalAmount) { // Add capture URL!
```
alexbishop1: Please remove the `// Add capture URL!` comment (that’s kind of my fault because I typed it on your machine!)

src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
```diff
@@ -62,6 +66,15 @@ private PaymentWithAllLinks(String chargeId, long amount, PaymentState state, St
...
         this.links.addKnownLinksValueOf(paymentConnectorResponseLinks);
     }
+    
+    private boolean isCapturePresent(List<PaymentConnectorResponseLink> paymentConnectorResponseLinks) {
+        for (PaymentConnectorResponseLink link : paymentConnectorResponseLinks) {
+            if(link.getRel().equals("capture")) {
+                return true;
+            }
+        }
+        return false;
```
alexbishop1: We can do something like this:

```java
return paymentConnectorResponseLinks.stream().anyMatch(link -> "capture".equals(link.getRel()));
```

Then we probably don’t even need to have a separate `isCapturePresent` method and can just inline it.

(Please also put the constant first when comparing — `"capture".equals(link.getRel())` rather than `link.getRel().equals("capture")` — because it won’t throw an exception if `link.getRel()` returns `null`.)

src/test/java/uk/gov/pay/api/model/PaymentTest.java
```diff
@@ -46,7 +46,7 @@ public void shouldNotPrintEmailAndCardDetailsWhenToString() throws IOException {
                 "  }\n" +
                 "}", ChargeFromResponse.class);
 
-        PaymentWithAllLinks payment = PaymentWithAllLinks.valueOf(paymentFromConnector, selfUri, eventsUri, cancelUri, refundsUri);
+        PaymentWithAllLinks payment = PaymentWithAllLinks.valueOf(paymentFromConnector, selfUri, eventsUri, cancelUri, refundsUri, null);
```
alexbishop1: In reality, the capture URL isn’t ever going to be `null`, so I think we should probably pass in a dummy `URI` object (like we do for the other URLs, which are created on lines 21 through 24)

src/test/java/uk/gov/pay/api/resources/PaymentsResourceCreatePaymentTest.java
```diff
@@ -124,6 +124,7 @@ private PaymentWithAllLinks aSuccessfullyCreatedPayment() {
                 URI.create(paymentUri + "/cancel"),
                 URI.create(paymentUri + "/refunds"),
                 null,
```
alexbishop1: As above, please do `URI.create(paymentUri + "/capture")`

src/test/java/uk/gov/pay/api/service/GetPaymentServiceTest.java
```diff
@@ -100,5 +102,17 @@ public void testGetPaymentWithCorporateCardSurcharge() {
         CardPayment payment = (CardPayment) paymentResponse.getPayment();
         assertThat(payment.getCorporateCardSurcharge().get(), is(250L));
         assertThat(payment.getTotalAmount().get(), is(2250L));
+        assertThat(paymentResponse.getLinks().getCapture(), is(nullValue()));
+    }
+
+    @Test
+    @PactVerification({"connector"})
+    @Pacts(pacts = {"publicapi-connector-get-payment-with-awaiting_capture_request_state"})
```
alexbishop1: Not sure about this mixing dashes and underscores in the name

swagger/swagger.json
```diff
@@ -799,9 +799,39 @@
...
+          "$ref" : "#/definitions/PostLink"
+        },
+        "capture" : {
+          "description" : "capture",
+          "readOnly" : true,
+          "$ref" : "#/definitions/PostLink"
         }
       },
-      "description" : "links for events resource"
+      "description" : "self,events and next links of a Payment"
```
alexbishop1: I wonder if maybe this should say some like “links for payment” so we don’t have to update it every time we add a new one

src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
SandorArpa: my bad :(

src/test/java/uk/gov/pay/api/service/GetPaymentServiceTest.java
SandorArpa: my bad!

src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
```diff
@@ -84,7 +88,8 @@ public static PaymentWithAllLinks valueOf(ChargeFromResponse paymentConnector,
                                               URI selfLink,
                                               URI paymentEventsUri,
                                               URI paymentCancelUri,
-                                              URI paymentRefundsUri) {
+                                              URI paymentRefundsUri,
+                                              URI captureUri) {
```
alexbishop1: Should have caught this earlier but it should probably be `paymentsCaptureUri` to match the others

src/test/java/uk/gov/pay/api/model/PaymentTest.java
alexbishop1: Looks like this comment was missed

src/test/java/uk/gov/pay/api/it/GetPaymentITest.java
```diff
@@ -319,4 +319,19 @@ private ValidatableResponse getPaymentEventsResponse(String bearerToken, String
                 .get(String.format("/v1/payments/%s/events", paymentId))
                 .then();
     }
+
+    @Test
+    public void getPayment_ReturnsPaymentWithCaptureUrl() {
```
alexbishop1: Could you move this test above the private methods please?

src/test/java/uk/gov/pay/api/it/GetPaymentITest.java
```diff
@@ -319,4 +319,19 @@ private ValidatableResponse getPaymentEventsResponse(String bearerToken, String
...
+        publicAuthMock.mapBearerTokenToAccountId(API_KEY, GATEWAY_ACCOUNT_ID);
+        connectorMock.respondWithChargeFoundAndCaptureUrl(250, GATEWAY_ACCOUNT_ID, CHARGE_ID, CAPTURED, RETURN_URL,
```
alexbishop1: It’s a bit unrealistic that the status is `CAPTURED` because such a payment shouldn’t return a capture URL. I have more thoughts on this, which I’ll leave on `ConnectorMockClient`.

src/test/java/uk/gov/pay/api/utils/mocks/ConnectorMockClient.java
```diff
@@ -244,6 +248,25 @@ public void respondWithChargeFound(long amount, String gatewayAccountId, String
                         .withBody(chargeResponseBody));
     }
 
+    public void respondWithChargeFoundAndCaptureUrl(long amount, String gatewayAccountId, String chargeId, PaymentState state, String returnUrl,
```
alexbishop1: Rather than adding a whole new method, we could instead modify the existing `respondWithChargeFound` so that if the status is `AWAITING_CAPTURE_REQUEST`, it adds a link with the capture URL. Then the test would just need to call `respondWithChargeFound` and pass in the appropriate status.

PR #281, opened 2018-11-05
PP-4297 add awaiting capture request single charge

SandorArpa:
## WHAT YOU DID
Make GetPaymentServiceTest pass the new pact test

## HOW
- write pact test for AWAITING_CAPTURE_REQUEST
- add/refactor process to handle capture uri
- update other pact tests to check if href is null when state is not eligible for capture


SandorArpa:
Duplicate

PR #282, opened 2018-11-06
PP-4297 Add capture URI to search results

SandorArpa:
## WHAT YOU DID
- add capture URI to search results
- add pact test

## How to test

When a charge is on AWAITING_CAPTURE_REQUEST state in connector, calling the endpoint /v1/payments should include

"results":[
....
  "_links":
    "capture": {
             "href": "xxxx/v1/payments/xxxxx/capture",
             "method": "POST"
         }
...]
in the response payload.
Any other states should include

"results":
  "_links":
    "capture": null

### DIFF
--- a/src/test/java/uk/gov/pay/api/it/PaymentResourceSearchITest.java
+++ b/src/test/java/uk/gov/pay/api/it/PaymentResourceSearchITest.java
@@ -27,6 +27,7 @@ import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static org.eclipse.jetty.http.HttpStatus.OK_200;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.Matchers.hasKey;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertThat;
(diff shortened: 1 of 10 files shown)


src/test/resources/pacts/publicapi-connector-search-payment-with-charge-in-awaiting-capture-state.json
```diff
@@ -0,0 +1,185 @@
...
+              "links": [
+                {
+                  "rel": "self",
+                  "method": "GET",
+                  "href": "https://connector/v1/api/accounts/123456/charges/ch_123abc456def"
+                },
+                {
+                  "rel": "refunds",
+                  "method": "GET",
+                  "href": "https://connector/v1/api/accounts/123456/charges/ch_123abc456def/refunds"
```
DanailMinchev: Looks like we are using different hosts for connector in one file, if you would like you can change it to one variant. They are ignored anyway, because we have matchers, but just to be consistent within one file

src/test/java/uk/gov/pay/api/service/CardPaymentSearchServiceTest.java
```diff
@@ -146,7 +147,23 @@ public void searchShouldReturnAResponseWithOneTransaction_whenFilteringByCardHol
...
+        Account account = new Account("123456", TokenPaymentType.CARD);
+        Response response =
+                paymentSearchService.doSearch(account, null, null,
+                        null, null, null,
+                        null, null, null,
+                        null, null, null, null);
+        JsonAssert.with(response.getEntity().toString())
+                .assertThat("results[0]._links", hasKey("capture"))
+                .assertThat("results[0]._links.capture.method", is("POST"));
+    }
```
alexbishop1: It would be nice to have a test to check we get {{"capture": null}} if the payment is not in the {{AWAITING CAPTURE REQUEST}} state

src/test/java/uk/gov/pay/api/it/PaymentResourceSearchITest.java
```diff
@@ -570,6 +571,35 @@ public void searchPayments_withMandateId_whenCardPayment_shouldReturnABadRequest
                 .assertThat("$.description", is("Invalid parameters: agreement_id. See Public API documentation for the correct data formats"));
     }
 
+    @Test
+    public void searchPayments_ShouldIncludeCaptureLink_whenReturnedFromConnector() {
+        publicAuthMock.mapBearerTokenToAccountId(API_KEY, GATEWAY_ACCOUNT_ID);
+        String SUBMITTED_STATE = "submitted";
```
alexbishop1: This isn’t a constant, so it should really be in regular `camelCase` rather than `SCREAMING_SNAKE_CASE`

src/test/java/uk/gov/pay/api/it/PaymentResourceSearchITest.java
```diff
@@ -570,6 +571,35 @@ public void searchPayments_withMandateId_whenCardPayment_shouldReturnABadRequest
...
+        connectorMock.respondOk_whenSearchCharges(GATEWAY_ACCOUNT_ID, TEST_REFERENCE, null, null, null, null, null, null, null, null,
+                payments
+        );
+
+        searchPayments(API_KEY, ImmutableMap.of("reference", TEST_REFERENCE))
+                .statusCode(200)
+                .contentType(JSON)
+                .body("results[0]._links", hasKey("capture"))
+                .body("results[0]._links.capture.method", is("POST"));
+
```
alexbishop1: We probably have enough information to assert on the `href` too

PR #296, opened 2018-11-08
Pp 4113 cleanup links

simad:
## WHAT YOU DID
- Removed refunds link from refund
- Renamed links to _links, payment_url to payment, amount_submitted to amount and charge_id to payment_id
- Updated docs

with @cobainc0


src/test/resources/pacts/publicapi-connector-search-refunds-with-from-and-to-date.json
```diff
@@ -148,28 +148,28 @@
                 }
               ]
             },
-            "$.results[0].links[0].href": {
+            "$.results[0]._links[0].href": {
```
heathd: did you really mean to change these? is the response structure from connector changing?

src/test/java/uk/gov/pay/api/service/SearchRefundsServiceTest.java
```diff
@@ -85,18 +82,20 @@ public void getAllRefundsShouldReturnCorrectTotalAndPageAndDisplaySize() {
                 .assertThat("$.results[0].charge_id", is(extChargeId))
                 .assertThat("$.results[0].amount_submitted", is(98))
                 .assertThat("$.results[0].status", is("available"))
-                .assertThat("$.results[0].links", not(hasKey("refunds")))
-                .assertThat("$.results[0].links.self.href", is(format("http://publicapi.test.localhost/v1/payments/%s/refunds/%s", extChargeId, refundId1)))
-                .assertThat("$.results[0].links.payment.href", is(format("http://publicapi.test.localhost/v1/payments/%s", extChargeId)))
+                .assertThat("$.results[0]", not(hasKey("links")))
+                .assertThat("$.results[0]._links", not(hasKey("refunds")))
```
heathd: same here, not sure this should change, because it's testing the response from the connector pact, right?

heathd: actually this is the external response from public api, so it should change

src/test/resources/pacts/publicapi-connector-search-refunds-with-from-and-to-date.json
simad: Ooops! Fixed, good catch 👍

PR #308, opened 2018-11-14
PP-4162 Make billing address Optional in CardDetails 

SandorArpa:
## WHAT YOU DID

- Made billing address optional on domain objects.
- Added tests for both the get payment and search payments endpoints that the billing_address property has a null value when no address is returned from connector.

with @stephencdaly

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CardDetails.java
+++ b/src/main/java/uk/gov/pay/api/model/CardDetails.java
@@ -5,6 +5,8 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 
+import java.util.Optional;
+
 import static com.fasterxml.jackson.annotation.JsonInclude.Include.ALWAYS;
 
 @JsonInclude(ALWAYS)
@@ -16,7 +18,7 @@ public class CardDetails {
 
     @JsonProperty("first_digits_card_number")
     private final String firstDigitsCardNumber;
-    
+
     @JsonProperty("cardholder_name")
     private final String cardHolderName;
 
@@ -30,7 +32,7 @@ public class CardDetails {
     private final String cardBrand;
 
     public CardDetails(@JsonProperty("last_digits_card_number") String lastDigitsCardNumber,
-                       @JsonProperty("first_digits_card_number") String firstDigitsCardNumber, 
+                       @JsonProperty("first_digits_card_number") String firstDigitsCardNumber,
                        @JsonProperty("cardholder_name") String cardHolderName,
                        @JsonProperty("expiry_date") String expiryDate,
                        @JsonProperty("billing_address") Address billingAddress,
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/api/it/PaymentResourceSearchITest.java
```diff
@@ -599,7 +601,30 @@ public void searchPayments_ShouldIncludeCaptureLink_whenReturnedFromConnector()
                 .body("results[0]._links", hasKey("capture"))
                 .body("results[0]._links.capture.method", is("POST"))
                 .body("results[0]._links.capture.href", is("http://publicapi.url/v1/payments/" + chargeId + "/capture"));
+    }
+    
+    @Test
+    public void searchPayments_getsResultsFromConnector_withNoBillingAddress() {
```
alexbishop1: Looks we don’t have a test for if there is a billing address asserting that one is there — might be useful to add one for completeness

src/test/java/uk/gov/pay/api/it/fixtures/PaymentResultBuilder.java
```diff
@@ -61,7 +61,7 @@ public CardDetails(uk.gov.pay.api.model.CardDetails cardDetails) {
             this.cardholder_name = cardDetails.getCardHolderName();
             this.expiry_date = cardDetails.getExpiryDate();
             this.card_brand = cardDetails.getCardBrand();
-            this.billing_address = cardDetails.getBillingAddress() == null ? null : new Address( cardDetails.getBillingAddress());
+            this.billing_address = cardDetails.getBillingAddress().isPresent() ? new Address( cardDetails.getBillingAddress().get()) : null;
```
alexbishop1:
```java
this.billing_address = cardDetails.getBillingAddress().map(Address::new).orElse(null);
```

src/main/java/uk/gov/pay/api/model/CardDetails.java
```diff
@@ -16,21 +18,22 @@
...
-    
+
     @JsonProperty("cardholder_name")
     private final String cardHolderName;
 
     @JsonProperty("expiry_date")
     private final String expiryDate;
 
     @JsonProperty("billing_address")
+    @JsonInclude(value = JsonInclude.Include.NON_EMPTY)
```
alexbishop1: So now I’m seeing the code, I think the story may have have gotten the requirements wrong here. I think we probably want to include the `"billing_address"` but have it be explicitly `null` to match the behaviour of the other properties inside `"card_details"` e.g. `"first_digits_card_number"` (which won’t be populated for older payments).

src/test/java/uk/gov/pay/api/it/PaymentResourceSearchITest.java
stephencdaly: This is already covered by `searchPayments_shouldOnlyReturnAllowedProperties`

src/test/java/uk/gov/pay/api/it/fixtures/PaymentResultBuilder.java
stephencdaly: done

src/test/java/uk/gov/pay/api/it/PaymentResourceSearchITest.java
alexbishop1: Right you are!

PR #309, opened 2018-11-14
PP-4162 Change how we build mock connector responses to deal with optionals

stephencdaly:
## WHAT YOU DID
- Our JsonStringBuilder class cannot handler optionals when building
  JSON strings.
- Refactored functionality of PaymentSearchResultBuilder into a base
  class and added PaymentSingleResultBuilder which also extends this
  and produces the JSON for a mocked single payment response from
  connector using Gson.
- Use PaymentSingleResultBuilder in ConnectorMockClient to build the
  mock connector response when we make a request to get a single payment.

with @SandorArpa

## How to test

- Does the refactoring and where we used it make sense?

### Logging


src/test/java/uk/gov/pay/api/utils/mocks/ConnectorMockClient.java
```diff
@@ -168,13 +168,10 @@ public void respondOk_whenCreateCharge(int amount, String gatewayAccountId, Stri
                                 settlementSummary,
                                 cardDetails,
                                 validGetLink(chargeLocation(gatewayAccountId, chargeId), "self"),
-                                validGetLink(nextUrl(chargeTokenId), "next_url"), validPostLink(nextUrlPost(), "next_url_post", "application/x-www-form-urlencoded",
-                                        new HashMap<String, String>() {{
-                                            put("chargeTokenId", chargeTokenId);
-                                        }}))));
+                                validGetLink(nextUrl(chargeTokenId), "next_url"),
+                                validPostLink(nextUrlPost(), "next_url_post", "application/x-www-form-urlencoded", getChargeIdTokenMap(chargeTokenId)))));
```
stephencdaly: This change is to not instantiate the map using the instance initializer, as this cannot be deserialized by JSON.
We've left a comment on the `getChargeIdTokenMap` method about this.

More info can be found here:
https://stackoverflow.com/questions/45876600/hashmap-to-gson-with-maps-as-values

src/test/java/uk/gov/pay/api/it/fixtures/PaymentResultBuilder.java
```diff
@@ -0,0 +1,228 @@
+package uk.gov.pay.api.it.fixtures;
```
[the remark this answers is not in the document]
alexbishop1: Public API responses will actually have `card_brand` in both places (it should be in `card_details` but we have to also keep it at the top level for backwards-compatibility)

src/test/java/uk/gov/pay/api/it/fixtures/PaymentResultBuilder.java
```diff
@@ -0,0 +1,228 @@
...
+
+    static {
+        states.add(new TestPaymentState("created", false));
+        states.add(new TestPaymentState("started", false));
+        states.add(new TestPaymentState("submitted", false));
+        states.add(new TestPaymentSuccessState("success"));
+    }
+
+    protected long amount = DEFAULT_AMOUNT;
+    protected String reference = null;
```
alexbishop1: Not sure why some of these are set to `null` and others aren’t but I see the old code did this too

PR #348, opened 2019-01-09
PP-3437: Add no auth token test

oswaldquek:
This replaces this e2e test:
https://github.com/alphagov/pay-endtoend/blob/3b6bab09049c4c712d897d1834060f7d527da125/src/test/java/uk/gov/pay/endtoend/tests/publicapi/AuthorisationIT.java

I've also taken the opportunity to clean up some test code.

Please also approve https://github.com/alphagov/pay-endtoend/pull/701!

@oswaldquek

### DIFF
--- a/src/test/java/uk/gov/pay/api/it/AuthorisationTest.java
+++ b/src/test/java/uk/gov/pay/api/it/AuthorisationTest.java
@@ -1,0 +1,36 @@
+package uk.gov.pay.api.it;
+
+import io.dropwizard.testing.junit.DropwizardAppRule;
+import org.apache.http.HttpStatus;
+import org.junit.Rule;
+import org.junit.Test;
+import uk.gov.pay.api.app.PublicApi;
+import uk.gov.pay.api.app.config.PublicApiConfig;
+
+import static com.jayway.restassured.RestAssured.given;
+import static com.jayway.restassured.http.ContentType.JSON;
+import static io.dropwizard.testing.ConfigOverride.config;
+import static io.dropwizard.testing.ResourceHelpers.resourceFilePath;
+import static uk.gov.pay.api.utils.Payloads.aSuccessfulPaymentPayload;
+
+public class AuthorisationTest {
+
+    @Rule
+    public DropwizardAppRule<PublicApiConfig> app = new DropwizardAppRule<>(
+            PublicApi.class,
+            resourceFilePath("config/test-config.yaml"),
+            config("connectorUrl", "http://localhost"),
+            config("connectorDDUrl", "http://localhost"),
+            config("publicAuthUrl", "http://localhost"));
+
+    @Test
+    public void shouldRefuseAuthorisationIfTokenIsNotPresent() {
+        given().port(app.getLocalPort())
+                .body(aSuccessfulPaymentPayload())
+                .accept(JSON)
+                .contentType(JSON)
+                .post("/v1/payments/")
+                .then()
+                .statusCode(HttpStatus.SC_UNAUTHORIZED);
+    }
+}
--- a/src/test/java/uk/gov/pay/api/it/PaymentResourceITestBase.java
+++ b/src/test/java/uk/gov/pay/api/it/PaymentResourceITestBase.java
@@ -17,6 +17,7 @@ import uk.gov.pay.api.utils.mocks.ConnectorMockClient;
 
 import static io.dropwizard.testing.ConfigOverride.config;
 import static io.dropwizard.testing.ResourceHelpers.resourceFilePath;
+import static uk.gov.pay.api.utils.Urls.paymentLocationFor;
 
 public abstract class PaymentResourceITestBase {
     //Must use same secret set int confiPaymentsResourceReferenceVgured test-config.xml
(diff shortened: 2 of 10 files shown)


src/test/java/uk/gov/pay/api/utils/Payloads.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.pay.api.utils;
+
+public class Payloads {
```
danworth: Perhaps a builder pattern with the defaults from the no args constructor would offer more flexibility, but no need for greater complexity if you don't feel its needed. Just a passing thought.

src/test/java/uk/gov/pay/api/it/AuthorisationTest.java
```diff
@@ -0,0 +1,36 @@
...
+import uk.gov.pay.api.app.PublicApi;
+import uk.gov.pay.api.app.config.PublicApiConfig;
+
+import static com.jayway.restassured.RestAssured.given;
+import static com.jayway.restassured.http.ContentType.JSON;
+import static io.dropwizard.testing.ConfigOverride.config;
+import static io.dropwizard.testing.ResourceHelpers.resourceFilePath;
+import static uk.gov.pay.api.utils.Payloads.aSuccessfulPaymentPayload;
+
+public class AuthorisationTest {
```
danworth: This is a nice easy to understand test 👍

src/test/java/uk/gov/pay/api/it/directdebit/pact/DirectDebitPaymentTest.java
```diff
@@ -4,31 +4,32 @@
...
 import static com.jayway.restassured.http.ContentType.JSON;
 import static io.dropwizard.testing.ConfigOverride.config;
 import static io.dropwizard.testing.ResourceHelpers.resourceFilePath;
 import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.core.Is.is;
+import static uk.gov.pay.api.utils.Payloads.aSuccessfulPaymentPayload;
+import static uk.gov.pay.api.utils.Urls.directDebitFrontendSecureUrl;
 
-public class DirectDebitPaymentTest extends PaymentResourceITestBase {
```
danworth: 👍 Looks like the only thing we were taking from the base class was `paymentLocationFor()` which I see you've now added here. I'm in favour of a little duplication to make tests easier to understand.

src/test/java/uk/gov/pay/api/it/directdebit/pact/DirectDebitPaymentTest.java
```diff
@@ -109,14 +110,8 @@ private ValidatableResponse postPaymentResponse(String bearerToken, String paylo
...
 
-    private static String paymentPayload(long amount, String returnUrl, String description, String reference, String email) {
-        return new JsonStringBuilder()
-                .add("amount", amount)
-                .add("reference", reference)
-                .add("email", email)
-                .add("description", description)
-                .add("return_url", returnUrl)
-                .build();
+    private String paymentLocationFor(String chargeId) {
```
danworth: I guess this could be moved into your new Urls class? Then we can still remove the `extends PaymentResourceITestBase` and have this method available to the masses.

danworth:
Just a thought, should there be a positive test in `AuthorisationTest` to complement the `shouldRefuse....`

src/test/java/uk/gov/pay/api/it/directdebit/pact/DirectDebitPaymentTest.java
oswaldquek: hmm yeah I'll do that

src/test/java/uk/gov/pay/api/it/directdebit/pact/DirectDebitPaymentTest.java
oswaldquek: the reason I've removed the `extends` is because there are a bunch of things in PaymentResourceITestBase that aren't used here. i think the "favor composition over inheritance" principle Trumps "do not repeat yourself". i'm going to move the `paymentLocationFor` into the Urls class as you've suggested :)

src/test/java/uk/gov/pay/api/utils/Payloads.java
oswaldquek: yeah i think that will make things a bit too complex for now

src/test/java/uk/gov/pay/api/it/directdebit/pact/DirectDebitPaymentTest.java
danworth: 👍 Completely agree

src/test/java/uk/gov/pay/api/it/PaymentResourceITestBase.java
```diff
@@ -77,32 +80,24 @@ private String publicAuthBaseUrl() {
...
     String frontendUrlFor(TokenPaymentType paymentType) {
         return "http://frontend_" + paymentType.toString().toLowerCase() + "/charge/";
     }
-
-    protected String directDebitFrontendSecureUrl() {
-        return "http://frontend_direct_debit/secure/";
-    }
-
+    
     String paymentEventsLocationFor(String chargeId) {
```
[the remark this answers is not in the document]
oswaldquek: yeah probably another PR...

PR #360, opened 2019-01-28
PP-4610 Swagger - examples for properties

kbottla:
## WHAT YOU DID
- Added examples for few API properties
- Fixed definition for Payment Events
- Tagger refund endpoints


kbottla:
Thanks @georgeracu. I have put annotations close to `JsonProperty` annotation which on getters in some places, and on private fields with other.

PR #361, opened 2019-01-29
PP-4610 Swagger - fix authorization

kbottla:
## WHAT YOU DID
- Fixes Authorization in code samples
   curl  `-H 'Authorization: API_KEY'` instead of `curl -H 'Authorisation: API_KEY'`
- Renamed few model names

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.api.model;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 import uk.gov.pay.commons.model.SupportedLanguage;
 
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
```diff
@@ -13,6 +14,7 @@
  * Note that this class does not actually do any validation itself:
  * do that yourself before constructing an instance of this object
  **/
+@ApiModel(value = "CreatePaymentRequest")
```
oswaldquek: what's the effect of putting this `@ApiModel` here?
why does the "agreement_id" in this class not appear in the swagger.json file but everything else ("amount", "description", "reference", "return_url" etc) does?

kbottla: `@ApiModel(value = "CreatePaymentRequest")` will create model with the same name `CreatePaymentRequest` in swagger instead of defaulting to `ValidCreatePaymentRequest`

`agreement_id` is excluded for now and we intend to include it when Direct Debit is ready for use
https://github.com/alphagov/pay-publicapi/blob/master/pom.xml#L433

oswaldquek: ah so it's excluded in the pom.xml.

PR #363, opened 2019-01-30
PP-4597 output timestamps in miliseconds

georgeracu:
## WHAT YOU DID
Output time in milliseconds
Code cleaup


src/test/java/uk/gov/pay/api/it/GetDirectDebitEventsTest.java
```diff
@@ -42,19 +42,19 @@
...
-    
+
     @Test
     @PactVerification({"direct-debit-connector", "publicauth"})
     @Pacts(pacts = {"publicapi-direct-debit-connector-get-events"})
     @Pacts(pacts = {"publicapi-publicauth"}, publish = false)
     public void getDirectDebitEvents() {
-        String requestPath = "/v1/events?to_date=2018-03-13T10:00:05Z&from_date=2018-03-13T10:00:03Z&display_size=100&page=1&agreement_id=1";
-        
+        String requestPath = "/v1/events?to_date=2018-03-13T10:00:05.000Z&from_date=2018-03-13T10:00:03.000Z&display_size=100&page=1&agreement_id=1";
```
alexbishop1: Not sure about this one: this in a input date/time (provided by service users) and for those won’t always have millisecond precision

src/test/resources/pacts/publicapi-direct-debit-connector-get-events.json
```diff
@@ -38,8 +38,8 @@
         "method": "GET",
         "path": "/v1/events",
         "query": {
-          "to_date": ["2018-03-13T10:00:05Z"],
-          "from_date": ["2018-03-13T10:00:03Z"],
+          "to_date": ["2018-03-13T10:00:05.000Z"],
+          "from_date": ["2018-03-13T10:00:03.000Z"],
```
alexbishop1: Again, not sure about changing these

src/test/java/uk/gov/pay/api/it/GetDirectDebitEventsTest.java
georgeracu: Fair point, it slipped through :)

src/test/resources/pacts/publicapi-direct-debit-connector-get-events.json
georgeracu: This change is required as we create this request inside `publicapi` and we call `connector` with it. Milliseconds are added in `DateTimeUtils#toUTCDateString` as we use this method everywhere to format ZonedDateTime. We do not receive this parameters from outside world :)

alexbishop1: Ah, OK, I misunderstood where it was coming from

PR #365, opened 2019-02-01
PP-4597 output timestamps in miliseconds

georgeracu:
## WHAT YOU DID
- Removed custom JSON deserializer and using a common one
- Bumped pay-java-common 
- Code cleanup

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging


src/main/java/uk/gov/pay/api/service/ConnectorUriGenerator.java
```diff
@@ -85,11 +85,9 @@ public String eventsURI(Account account, Optional<ZonedDateTime> toDate, Optiona
 
         Map<String, String> params = new LinkedHashMap<>();
 
-        if (toDate.isPresent())
-            params.put("to_date", DateTimeUtils.toUTCDateString(toDate.get()));
+        toDate.ifPresent(zonedDateTime -> params.put("to_date", DateTimeUtils.toUTCDateString(zonedDateTime)));
```
alexbishop1: I think we can just use the formatter directly here:

```java
toDate.ifPresent(zonedDateTime -> params.put(ISO_INSTANT_MILLISECOND_PRECISION.format(zonedDateTime));
```

georgeracu: That's well spotted. We can remove this wrapper method and use the formatter directly :)

src/main/java/uk/gov/pay/api/service/ConnectorUriGenerator.java
```diff
@@ -85,11 +85,9 @@ public String eventsURI(Account account, Optional<ZonedDateTime> toDate, Optiona
 
         Map<String, String> params = new LinkedHashMap<>();
 
-        if (toDate.isPresent())
-            params.put("to_date", DateTimeUtils.toUTCDateString(toDate.get()));
+        toDate.ifPresent(zonedDateTime -> params.put("to_date", ISO_INSTANT_MILLISECOND_PRECISION.format(zonedDateTime)));
```
alexbishop1: Just realised we can do this!

```java
toDate.map(ISO_INSTANT_MILLISECOND_PRECISION::format).ifPresent(timestamp -> params.put("from_date", timestamp));
```

georgeracu: 👍 done

PR #376, opened 2019-02-12
Intellij fixups

rhowe-gds:
Address some things IntelliJ didn't like the look of.

Most of the attention was on `RedisContainer` where the `get()` from the `Map` of port mappings could throw a null. Add some guard code around that.

Remove some copypasta in `ConnectorMockClient` when constructing exceptions - we don't need `String.format()` here because we're passing a fixed string.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
@@ -34,7 +34,7 @@ public class ValidCreatePaymentRequest {
     private Boolean delayedCapture;
 
     public ValidCreatePaymentRequest(CreatePaymentRequest createPaymentRequest) {
-        amount = Objects.requireNonNull(createPaymentRequest.getAmount());
+        amount = createPaymentRequest.getAmount();
         reference = Objects.requireNonNull(createPaymentRequest.getReference());
         description = Objects.requireNonNull(createPaymentRequest.getDescription());
 
--- a/src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
+++ b/src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
@@ -14,13 +14,12 @@ import org.slf4j.LoggerFactory;
 import redis.clients.jedis.Jedis;
 
 import java.io.IOException;
-import java.util.List;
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
```diff
@@ -34,7 +34,7 @@
     private Boolean delayedCapture;
 
     public ValidCreatePaymentRequest(CreatePaymentRequest createPaymentRequest) {
-        amount = Objects.requireNonNull(createPaymentRequest.getAmount());
+        amount = createPaymentRequest.getAmount();
```
alexbishop1: This change confused me at first but then I realised that `createPaymentRequest.getAmount()` returns an `int`, which cannot be null

src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -54,23 +53,33 @@ public RedisContainer(DockerClient docker, String host) throws DockerException,
...
+                    .networkSettings()
+                    .ports())
+                    .get(redisPortSpec)
+                    .get(0);
+            logger.info("Redis host port: {}", portBinding.hostPort());
+            return parseInt(portBinding.hostPort());
+        } catch (NullPointerException e) {
+            logger.error("Unable to find host port mapping for {} in container {}", redisPortSpec, containerInfo.id());
+            throw e;
+        }
```
alexbishop1:
```java
        String redisPortSpec = INTERNAL_PORT + "/tcp";
        return Optional.ofNullable(containerInfo.networkSettings().ports())
                .map(ports -> ports.get(redisPortSpec))
                .map(redisPorts -> redisPorts.get(0))
                .map(PortBinding::hostPort)
                .map(Integer::parseInt).orElseThrow(() -> {
                    logger.error("Unable to find host port mapping for {} in container {}", redisPortSpec, containerInfo.id());
                    return new NullPointerException();
                });
```

src/main/java/uk/gov/pay/api/model/ValidCreatePaymentRequest.java
rhowe-gds: As per the corresponding commit message ;)

PR #410, opened 2019-04-05
Firebreak - Swagger validation

kbottla:
## What you did
Explorer Swagger request validator library to validate swagger.json file (generated through annotation) with expected/actual responses https://bitbucket.org/atlassian/swagger-request-validator

Swagger can be validated in multiple ways with the library. Below are the ones that are close to frameworks we use / independent of frameworks (also variants includes using springmvc, mockmvc, spring web client)

### **1. Simple request/response validation**
Validates swagger.json with expected responses (json)
 
 **Pros**
 - Quick feedback
 - Independent of framework

 **Cons**
 - Can't actually test if swagger file is incorrect. An incorrect swagger with an incorrect expected responses will still pass
### **2. Pact provider validation**

Validates swagger.json with a pact interaction (local file / from pact-broker)

**Pros**
- Quick feedback

**Cons**
- Only useful when am actual consumer generates pact interactions. Generating interactions ourselves is not useful
- Can't actually test if swagger file is incorrect. Am incorrect swagger with an incorrect expected responses will still pass

### **3. Rest assured**

Validates swagger.json with actual responses to publicapi application

**Pros**
- Invokes requests on application which can detect if swagger itself is invalid

**Cons**
- Takes longer to run but shouldn't be a deal break

## What can it detect

Validator can detect the following in both swagger.json / responses (mock / actual responses)

- Schema mismatches ( objects, fields, datatype, enumerate values..)
- Unexpected fields (additional fields in swagger but not in response and vice verse)
- Request/Response headers

_Can also whitelist any validation errors if it is persistent issue that can't be fixed / to be ignored_ 

## Any issues

- Encountered issues with dependencies especially com.github.fge.json-schema-validator as it conflicts with another to read JsonSchema. Excluding this library seems fine but may have impacted existing integration tests using mockserver 
- Seems to work fine on both Java 8 & Java 11, but pacts provider tests doesn't work the latest (3.6.2) pact library.

## Additional info

Swagger request validator is listed as open source tool for Java by Swagger.io https://swagger.io/tools/open-source/open-source-integrations/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -169,6 +169,12 @@
             <artifactId>mockserver-netty</artifactId>
             <version>${mockserver.version}</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.github.fge</groupId>
+                    <artifactId>json-schema-validator</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.mock-server</groupId>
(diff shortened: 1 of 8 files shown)


kbottla:
Closing for now but can be referenced if we want to add swagger validation

PR #415, opened 2019-04-08
PP-5053 Upgrade to Java 11

danworth:
## WHAT YOU DID
This upgrades PublicApi to Java 11. Note that `JAVA_HOME` was previously set to JDK11 in the `Jenkins` file so no need to do it here.


danworth:
I forgot to update the trust store location... @kbottla @alexbishop1

PR #434, opened 2019-04-26
PP-4886: Refactor RequestJsonParser

oswaldquek:
Extract validations into its own method. This makes things clearer.

### DIFF
--- a/src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
+++ b/src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
@@ -2,11 +2,11 @@ package uk.gov.pay.api.json;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.apache.http.HttpStatus;
 import uk.gov.pay.api.exception.BadRequestException;
 import uk.gov.pay.api.model.CreatePaymentRefundRequest;
 import uk.gov.pay.api.model.CreatePaymentRequest;
 import uk.gov.pay.api.model.PaymentError;
+import uk.gov.pay.api.model.PaymentError.Code;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
```diff
@@ -41,136 +43,123 @@
...
+            throw new WebApplicationException(Response.status(SC_UNPROCESSABLE_ENTITY).entity(paymentError).build());
         }
-        return fieldValue;
+        return metadata;
     }
-
-    private static boolean notNull(JsonNode fieldNode) {
-        return fieldNode != null && !fieldNode.isNull();
+    
+    private static <T> T getValue(JsonNode jsonNode,
```
danworth: Perhaps calling this `validateAndGetValue` might be more descriptive and in-line with other private methods since it is validating as part of return the value.

src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
```diff
@@ -41,136 +43,123 @@
...
-        );
-        return fieldValue;
+    private static String validateAndGetReference(JsonNode paymentRequest) {
+        return getStringValue(
+                paymentRequest.get(REFERENCE_FIELD_NAME),
+                aPaymentError(REFERENCE_FIELD_NAME, CREATE_PAYMENT_VALIDATION_ERROR,"Must be a valid string format"),
+                aPaymentError(REFERENCE_FIELD_NAME, CREATE_PAYMENT_MISSING_FIELD_ERROR));
+    }
+    
+    private static String getStringValue(JsonNode jsonNode, PaymentError validationError, PaymentError missingError) {
```
danworth: perhaps calling this `validateAndGetString` might be more descriptive of what its doing and consistent with other private method names.

PR #437, opened 2019-04-29
PP-4887: Refactor to use builders

oswaldquek:


src/test/java/uk/gov/pay/api/utils/mocks/ChargeResponseFromConnector.java
```diff
@@ -0,0 +1,198 @@
...
+    public final String chargeId, returnUrl, description, reference, email, paymentProvider, gatewayTransactionId, createdDate;
+    public final SupportedLanguage language; 
+    public final boolean delayedCapture; 
+    public final RefundSummary refundSummary; 
+    public final SettlementSummary settlementSummary; 
+    public final CardDetails cardDetails;
+    public final List<Map<?, ?>> links;
+    public final Optional<Map<String, Object>> metadata;
+
+    private ChargeResponseFromConnector(long amount, String chargeId, PaymentState state, String returnUrl,
```
danworth: Sure there's benefits either way but since this constructor is private and can only be created via a builder then it could just take the instance of the builder as the argument rather than having to deconstruct it and have lots of arguments? e.g. `private ChargeResponseFromConnector(ChargeResponseFromConnectorBuilder builder)`. This is just my opinion and I would accept it as-is if you prefer.

src/test/java/uk/gov/pay/api/utils/mocks/ChargeResponseFromConnector.java
```diff
@@ -0,0 +1,198 @@
...
+import uk.gov.pay.commons.model.SupportedLanguage;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+public class ChargeResponseFromConnector {
+    public final Long amount, corporateCardSurcharge, totalAmount;
```
danworth: I get that this is only within test but I think we agreed no public immutable fields.

https://docs.google.com/document/d/1iWzWIsitOUdlyBZGqWzT49R27OqHidxl8Kh5-FMcrSY/edit

oswaldquek: yup good call, fixed.

src/test/java/uk/gov/pay/api/utils/mocks/ChargeResponseFromConnector.java
```diff
@@ -0,0 +1,267 @@
...
+
+    public List<Map<?, ?>> getLinks() {
+        return links;
+    }
+
+    public Optional<Map<String, Object>> getMetadata() {
+        return metadata;
+    }
+
+    private ChargeResponseFromConnector(ChargeResponseFromConnectorBuilder builder) {
```
danworth: I'd be inclined to put the constructor before the getters which might help cut down the impact on readability. Also I'm not sure we discussed `public` fields purely in the context of test code so perhaps worth raising again at some point if you feel it worth a challenge (my only concern is it sets a precedent which someone might miss-interpret and they find their way into `main`).

oswaldquek: doh. yeah i agree but i don't really want to wait for another long build time. all the more reason to move to kotlin with its implicit getters or relax the use `public final` fields in test classes because it's just code pollution.

src/test/java/uk/gov/pay/api/utils/mocks/ChargeResponseFromConnector.java
```diff
@@ -0,0 +1,267 @@
...
+        this.createdDate = builder.createdDate;
+        this.language = builder.language;
+        this.delayedCapture = builder.delayedCapture;
+        this.corporateCardSurcharge = builder.corporateCardSurcharge;
+        this.totalAmount = builder.totalAmount;
+        this.refundSummary = builder.refundSummary;
+        this.settlementSummary = builder.settlementSummary;
+        this.cardDetails = builder.cardDetails;
+        this.links = builder.links;
+        this.metadata = builder.metadata == null || builder.metadata.isEmpty() ? Optional.empty() : Optional.of(builder.metadata);;
```
danworth: Just realised that the Optional wrapping can move to the getter now if you'd prefer to keep the constructor simpler.

PR #447, opened 2019-05-01
PP-4886: Fix error description and multiple errors

oswaldquek:
The error description should begin with "Invalid attribute value: metadata.".
Where there are multiple they should be delimited with a period, not semicolon+newline.


oswaldquek:
@alexbishop1 I've addressed your comments. Regarding possibly changing the messages from connector, I think we should treat connector as its own independent service, so it should return what went wrong (by specifying it was the metadata field that failed validation in this case).

PR #462, opened 2019-05-09
PP-4886: Test for empty metadata

oswaldquek:
Test for where metadata supplied when creating a payment is

```
"metadata":{}
```


src/test/java/uk/gov/pay/api/it/CreatePaymentITest.java
```diff
@@ -51,6 +52,31 @@
     private static final String SUCCESS_PAYLOAD = paymentPayload(AMOUNT, RETURN_URL, DESCRIPTION, REFERENCE);
     private static final String GATEWAY_TRANSACTION_ID = "gateway-tx-123456";
 
+    @Test
+    public void createCardPaymentWithEmptyMetadata() {
```
alexbishop1: Could this maybe be called something like `createCardPaymentWithEmptyMetadataDoesNotStoreMetadata` so the name summarises the both what happens and the expected result?

PR #463, opened 2019-05-09
Bau use wiremock

oswaldquek:
This PR makes https://github.com/alphagov/pay-publicapi/pull/435 irrelevant. I find the API for wiremock is much easier to use/understand than the mockserver ones. 

A major change in this PR is that previously, when mocking calls to an external service (publicAuth, connector, ddConnector) the exact request and response was set on the mockServer class. I've removed the "exact request" bit in this PR so when writing tests, only the expected Url and responses are specified. This is ok because we already verify the wiremock server is called with certain json bodies.

Using wiremock causes the test run to reduce from 5 minutes to 2 minutes. This caused many test failures due to the redis rate limiting in integration tests. To rectify this I've created a method to clear the redis cache which can be called before a test run.

Also some miscellaneous code cleanups flagged by intellij.

Unfortunately I couldn't make this a smaller commit, sorry!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
         <dropwizard.version>1.3.10</dropwizard.version>
         <guice.version>4.2.2</guice.version>
         <guava.version>27.1-jre</guava.version>
-        <mockserver.version>5.1.0</mockserver.version>
+        <wiremock.version>2.23.2</wiremock.version>
         <swagger.jersey2.version>1.5.22</swagger.jersey2.version>
         <hamcrest.version>2.1</hamcrest.version>
         <jackson.version>2.9.8</jackson.version>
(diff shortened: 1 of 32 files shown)


pom.xml
```diff
@@ -164,15 +164,31 @@
...
-            <groupId>org.mock-server</groupId>
-            <artifactId>mockserver-netty</artifactId>
-            <version>${mockserver.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.mock-server</groupId>
-            <artifactId>mockserver-client-java</artifactId>
-            <version>${mockserver.version}</version>
+            <groupId>com.github.tomakehurst</groupId>
```
rhowe-gds: Should this be `<scope>test</scope>`?

src/test/java/uk/gov/pay/api/it/ResourcesFilterITestBase.java
```diff
@@ -1,65 +1,104 @@
...
+
+    @ClassRule
+    public static WireMockClassRule publicAuthMock = new WireMockClassRule(PUBLIC_AUTH_PORT);
+
+    @Rule
+    public DropwizardAppRule<PublicApiConfig> app = new DropwizardAppRule<>(
+            PublicApi.class, 
+            resourceFilePath("config/test-config.yaml"), 
+            config("connectorUrl", "http://localhost:" + CONNECTOR_PORT), 
+            config("connectorDDUrl", "http://localhost"),
```
rhowe-gds: Does this not need a port number and/or path?

oswaldquek: it doesn't because the classes that extend it don't use dd connector. i find setting unneeded dependencies to `http://localhost` useful because it gives us information that this test doesn't depend on dd connector.

rhowe-gds: Ah, `http://unused-by-this-test.invalid` might be clearer but fine

PR #465, opened 2019-05-10
PP-5100 Remove validation from CreatePaymentRequestDeserialiser

danworth:
- As step towards refactoring to use Java Bean Validation rather than custom validation
methods this moves the validation of the CreatePaymentRequest outside of the
deserialisation and temporarily into the Resource. The important part is the removal of `validator.validate(paymentRequest);` from within `CreatePaymentRequestDeserializer`
- When refactoring is complete the validation will be invoked via the `@Valid`
annotation on the method parameter and the `PaymentRequestValidator` can be removed
entirely.
- Removed validation related tests from `CreatePaymentRequestDeserializerTest`
which are covered by validation tests within `PaymentRequestValidatorTest`
```
deserialize_shouldThrowValidationException_whenAmountIsLessThanMinimum = validateMinimumAmount_shouldFailValue
deserialize_shouldThrowValidationException_whenAmountIsMoreThanMaximum = validateMaximumAmount_shouldFailValue
deserialize_shouldThrowValidationException_whenReturnUrlLengthIsMoreThan2000CharactersLength = validateReturnUrlMaxLength_shouldFailValue
deserialize_shouldThrowValidationException_whenReturnUrlIsAMalformedUrl = validateReturnUrlNotHttps_shouldFailValue
deserialize_shouldThrowValidationException_whenReferenceIsMoreThan255CharactersLength = validateReferenceMaxLength_shouldFailValue
deserialize_shouldThrowValidationException_whenDescriptionIsMoreThan255CharactersLength = validateDescriptionMaxLength_shouldFailValue
deserialize_shouldThrowValidationException_whenAgreementIdIsMoreThan26CharactersLength = validateAgreementIdMaxLength_shouldFailValue
deserialize_shouldThrowValidationException_whenLanguageIsNotSupported = validateUnsupportedLanguage_shouldFailValue
deserialize_shouldThrowValidationException_whenEmailIs255Character = validateEmailMaxLength_shouldFailValue
deserialize_shouldThrowValidationException_whenCardholderNameIs256Character = validateCardHolderNameMaxLength_shouldFailValue
deserialize_shouldThrowValidationException_whenLine1Is256Character = validateLine1MaxLength_shouldFailValue
deserialize_shouldThrowValidationException_whenPostcodeIs26Character = validatePostCodeMaxLength_shouldFailValue
deserialize_shouldThrowValidationException_whenCityIs256Character = validateCityMaxLength_shouldFailValue
deserialize_shouldThrowValidationException_whenCountryIsLessThan2Character = validateCountryMinLength_shouldFailValue
deserialize_shouldNotThrowValidationException_whenCountryIsEmptyString = validateCountryEmpty_shouldPass

```
## WHAT YOU DID
Please see commit notes. Essentially separate the deserialization from the validation as a prerequisite to refactoring the validation to use java validation spec. There should be no change to any responses.

### DIFF
--- a/src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
+++ b/src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
@@ -15,6 +15,7 @@ import org.slf4j.LoggerFactory;
 import uk.gov.pay.api.auth.Account;
 import uk.gov.pay.api.exception.CaptureChargeException;
 import uk.gov.pay.api.exception.GetEventsException;
+import uk.gov.pay.api.model.CreatePaymentRequest;
 import uk.gov.pay.api.model.CreatePaymentResult;
 import uk.gov.pay.api.model.PaymentError;
 import uk.gov.pay.api.model.PaymentEvents;
@@ -30,8 +31,10 @@ import uk.gov.pay.api.service.CreatePaymentService;
 import uk.gov.pay.api.service.GetPaymentService;
 import uk.gov.pay.api.service.PaymentSearchService;
 import uk.gov.pay.api.service.PublicApiUriGenerator;
+import uk.gov.pay.api.validation.PaymentRequestValidator;
 
 import javax.inject.Inject;
+import javax.validation.Valid;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
@@ -64,6 +67,7 @@ public class PaymentsResource {
     private final GetPaymentService getPaymentService;
     private final CapturePaymentService capturePaymentService;
     private final CancelPaymentService cancelPaymentService;
+    private PaymentRequestValidator paymentRequestValidator;
 
     @Inject
     public PaymentsResource(Client client,
@@ -73,7 +77,8 @@ public class PaymentsResource {
                             ConnectorUriGenerator connectorUriGenerator,
                             GetPaymentService getPaymentService,
                             CapturePaymentService capturePaymentService,
-                            CancelPaymentService cancelPaymentService) {
+                            CancelPaymentService cancelPaymentService,
+                            PaymentRequestValidator paymentRequestValidator) {
         this.client = client;
         this.createPaymentService = createPaymentService;
         this.publicApiUriGenerator = publicApiUriGenerator;
@@ -82,6 +87,7 @@ public class PaymentsResource {
         this.getPaymentService = getPaymentService;
         this.capturePaymentService = capturePaymentService;
         this.cancelPaymentService = cancelPaymentService;
+        this.paymentRequestValidator = paymentRequestValidator;
     }
 
     @GET
@@ -241,7 +247,11 @@ public class PaymentsResource {
             @ApiResponse(code = 429, message = "Too many requests", response = ApiErrorResponse.class),
             @ApiResponse(code = 500, message = "Downstream system error", response = PaymentError.class)})
     public Response createNewPayment(@ApiParam(value = "accountId", hidden = true) @Auth Account account,
-                                     @ApiParam(value = "requestPayload", required = true) ValidCreatePaymentRequest validCreatePaymentRequest) {
+                                     @ApiParam(value = "requestPayload", required = true) CreatePaymentRequest createPaymentRequest) {
+        logger.info("Payment create request parsed to {}", createPaymentRequest);
+
+        paymentRequestValidator.validate(createPaymentRequest);
+        ValidCreatePaymentRequest validCreatePaymentRequest = new ValidCreatePaymentRequest(createPaymentRequest);
         logger.info("Payment create request passed validation and parsed to {}", validCreatePaymentRequest);
 
         PaymentWithAllLinks createdPayment = createPaymentService.create(account, validCreatePaymentRequest);
--- a/src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
+++ b/src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
@@ -67,7 +67,7 @@ public class PaymentsResource {
     private final GetPaymentService getPaymentService;
     private final CapturePaymentService capturePaymentService;
     private final CancelPaymentService cancelPaymentService;
-    private PaymentRequestValidator paymentRequestValidator;
+    private final PaymentRequestValidator paymentRequestValidator;
 
     @Inject
     public PaymentsResource(Client client,
--- a/src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
+++ b/src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
@@ -14,6 +14,7 @@ import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
 import uk.gov.pay.api.exception.BadRequestException;
 import uk.gov.pay.api.model.Address;
+import uk.gov.pay.api.model.CreatePaymentRequest;
 import uk.gov.pay.api.model.ValidCreatePaymentRequest;
 import uk.gov.pay.api.validation.PaymentRequestValidator;
 import uk.gov.pay.api.validation.URLValidator;
@@ -43,7 +44,7 @@ public class CreatePaymentRequestDeserializerTest {
     @Before
     public void setup() {
         URLValidator urlValidator = URLValidator.urlValidatorValueOf(true);
-        deserializer = new CreatePaymentRequestDeserializer(new PaymentRequestValidator(urlValidator));
+        deserializer = new CreatePaymentRequestDeserializer();
     }
 
     @Test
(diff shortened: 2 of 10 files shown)


src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -64,6 +67,7 @@
     private final GetPaymentService getPaymentService;
     private final CapturePaymentService capturePaymentService;
     private final CancelPaymentService cancelPaymentService;
+    private PaymentRequestValidator paymentRequestValidator;
```
alexbishop1: Looks like this can be `final`

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -241,7 +247,11 @@ public Response searchPayments(@ApiParam(value = "accountId", hidden = true)
             @ApiResponse(code = 429, message = "Too many requests", response = ApiErrorResponse.class),
             @ApiResponse(code = 500, message = "Downstream system error", response = PaymentError.class)})
     public Response createNewPayment(@ApiParam(value = "accountId", hidden = true) @Auth Account account,
-                                     @ApiParam(value = "requestPayload", required = true) ValidCreatePaymentRequest validCreatePaymentRequest) {
+                                     @ApiParam(value = "requestPayload", required = true) CreatePaymentRequest createPaymentRequest) {
+        logger.info("Payment create request parsed to {}", createPaymentRequest);
+
+        paymentRequestValidator.validate(createPaymentRequest);
+        ValidCreatePaymentRequest validCreatePaymentRequest = new ValidCreatePaymentRequest(createPaymentRequest);
```
alexbishop1: Out of interest, have we decided yet if we’re planning to keep `ValidCreatePaymentRequest` in the long term? On the one hand, it’s great to have specific type for a validated request (it’s almost [typestate](https://en.wikipedia.org/wiki/Typestate_analysis)!) but it might be a bit redundant when we have `@Valid` and can basically assume that any request that survived deserialisation into a `CreatePaymentRequest` is valid.

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -241,7 +247,11 @@ public Response searchPayments(@ApiParam(value = "accountId", hidden = true)
...
         logger.info("Payment create request passed validation and parsed to {}", validCreatePaymentRequest);
```
alexbishop1: It’s possibly a bit redundant to include the `validCreatePaymentRequest` as now we have logging a few lines above that outputs the `createPaymentRequest`.

src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
```diff
@@ -624,19 +504,18 @@ public void shouldDeserializeARequestWithPrefilledCardholderDetailsSuccessfully(
...
-        assertThat(paymentRequest.getPrefilledCardholderDetails().isPresent(), is(true));
-        assertThat(paymentRequest.getPrefilledCardholderDetails().get().getCardholderName().isPresent(), is(true));
-        assertThat(paymentRequest.getPrefilledCardholderDetails().get().getCardholderName().get(), is("J Bogs"));
-        assertThat(paymentRequest.getPrefilledCardholderDetails().get().getBillingAddress().isPresent(), is(true));
-        Address billingAddress = paymentRequest.getPrefilledCardholderDetails().get().getBillingAddress().get();
+        assertThat(paymentRequest.getReturnUrl(), is("https://somewhere.gov.uk/rainbow/1"));
+        assertThat(paymentRequest.getLanguage(), is(nullValue()));
+        assertThat(paymentRequest.getDelayedCapture(), is(nullValue()));
+        assertThat(paymentRequest.getEmail(), is("j.bogs@example.org"));
+        assertThat(paymentRequest.getPrefilledCardholderDetails().getCardholderName().isPresent(), is(true));
```
alexbishop1: It’s… _interesting_ that we’ve ended up with an object graph where top-level getters like `getLanguage()` can return null but deeper getters like `getCarholderName()` return `Optional`. It would be nice to make this consistent.

src/test/java/uk/gov/pay/api/resources/PaymentsResourceCreatePaymentTest.java
```diff
@@ -76,22 +84,24 @@ public void setup() {
...
                 .returnUrl("https://somewhere.test")
                 .reference("my_ref")
                 .description("New Passport")
-                .build());
+                .build();
+
         PaymentWithAllLinks injectedResponse = aSuccessfullyCreatedPayment();
 
-        when(createPaymentService.create(account, createPaymentRequest)).thenReturn(injectedResponse);
+        when(createPaymentService.create(eq(account), ArgumentMatchers.any())).thenReturn(injectedResponse);
```
alexbishop1: Can we really not do better than `any()`? (Even `any(CreatePaymentRequest.class)` would be better!)

Feel free to add an `equals(…)` method to `CreatePaymentRequest` if it makes this sort of matching easier.

src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java
```diff
@@ -49,7 +50,10 @@ public void validParameters_withWelshLanguage_shouldSuccessValue() {
     @Test
     public void validateUnsupportedLanguage_shouldFailValue() {
         CreatePaymentRequest createPaymentRequest = createPaymentRequestBuilderWithReturnUrl().language("unsupported language").build();
+
         expectedException.expect(PaymentValidationException.class);
+        expectedException.expect(aValidationExceptionContaining("P0102", "Invalid attribute value: language. Must be \"en\" or \"cy\""));
```
alexbishop1: Do we really need both `expectedException.expect(…)` lines?

src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java
```diff
@@ -75,47 +82,160 @@ public void validateMaximumAmount_shouldSuccessValue() {
...
+        expectedException.expect(PaymentValidationException.class);
+        expectedException.expect(aValidationExceptionContaining("P0102", "Invalid attribute value: country. Must be exactly 2 characters length"));
+
+        paymentRequestValidator.validate(createPaymentRequest);
+    }
+    
+    @Test
+    public void validateCountryEmpty_shouldPass() {
+        CreatePaymentRequest createPaymentRequest = createPaymentRequestBuilderWithAgreementId().country("").build();
+
```
alexbishop1: It’s great that we’re reducing duplication of tests that test basically the same validation logic at different levels but it seems the functionality covered by the old `deserialize_shouldThrowValidationException_whenReturnUrlIsAMalformedUrl` test has been lost in the migration.

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
danworth: yes good catch, thanks.

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
danworth: Happy to discuss, my intention was to remove it since as you say its superfluous once we have validation on  `CreatePaymentRequest`, it also follows our pattern in Connector to have just one request type which is validated rather than parsing into a second.

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
danworth: I kept it temporarily whilst during the transition so if anything goes wrong its easier to spot whether it occurred within the custom deserialisation or the custom validation class. I agree that once we're done we should remove it to one log line.

src/test/java/uk/gov/pay/api/json/CreatePaymentRequestDeserializerTest.java
danworth: Yes, I'll fix this up as part of this story. Probably towards the end once the restructuring of validation is complete to avoid too much change on change.

src/test/java/uk/gov/pay/api/resources/PaymentsResourceCreatePaymentTest.java
danworth: Yes good point. I've added an equals method to `ValidCreatePaymentRequest` and used 
`        final ValidCreatePaymentRequest validCreatePaymentRequest = new ValidCreatePaymentRequest(createPaymentRequest);
`
then....
`eq(validCreatePaymentRequest)`

src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java
danworth: Hmmm partly I wanted to minimise the scope of the refactoring to aid review and minimise chances of introducing an error. However, looking at `aValidExceptionContaining` its a type safe matcher based upon `PaymentValidationException`, so you're right, we don't! I'll remove.

src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java
danworth: Great catch, thank you reviewing so diligently its much appreciated! I've added this in.

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
alexbishop1: Yeah, that sounds reasonable.

src/test/java/uk/gov/pay/api/resources/PaymentsResourceCreatePaymentTest.java
```diff
@@ -76,22 +84,26 @@ public void setup() {
...
                 .description("New Passport")
-                .build());
+                .build();
+
+        final ValidCreatePaymentRequest validCreatePaymentRequest = new ValidCreatePaymentRequest(createPaymentRequest);
+
         PaymentWithAllLinks injectedResponse = aSuccessfullyCreatedPayment();
 
-        when(createPaymentService.create(account, createPaymentRequest)).thenReturn(injectedResponse);
+        when(createPaymentService.create(eq(account), eq(validCreatePaymentRequest))).thenReturn(injectedResponse);
```
alexbishop1: Don’t need the `eq(…)` matchers if all arguments have them.

PR #471, opened 2019-05-13
PP 5100 Step towards using `@Valid` for validation of CreatePaymentRequest

danworth:
## WHAT YOU DID
To avoid large PR which would be harder to assure ourselves that behaviour hasn't changed this is another step towards replacing our validation approach with using `@Valid` java bean validation.

First commit
```
PP-5100 Introduce bean validation to CreateChargeRequest
    - Step towards removing validation from ValidateCreatePaymentRequest and to
    annotation based approach on CreatePaymentRequest. This now uses bean validation
    for `agreementId`.
    - Create ViolationExceptionMapper to create a PaymentError from the ConstraintViolation
    which will look for the api attribute name if it has a JsonProperty annotation.
    - As per existing PublicApi behaviour the ViolationExceptionMapper will return only
    the first exception, this can be changed later if we wish.
    - Add `@Valid` to the PaymentResource to begin validating the CreatePaymentRequest
    - Remove unit test PaymentRequestValidatorTest since this is covered within PaymentResourceAgreementIdValidationTest
```
Then building on this to move over some other basic attributes...
```
PP-5100 Migrate validation for create payment fields to @Valid strategy
    
    - Migrate amount, description, reference to validation under @Valid strategy
    - Remove from PaymentRequestValidator and associated tests since it is covered
    within PaymentsResourceAmountValidationTest
```

### DIFF
--- a/src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
@@ -1,0 +1,67 @@
+package uk.gov.pay.api.exception.mapper;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.dropwizard.jersey.validation.JerseyViolationException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.api.model.PaymentError;
+
+import javax.annotation.Priority;
+import javax.validation.ConstraintViolation;
+import javax.validation.Path;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.Optional;
+
+import static uk.gov.pay.api.model.PaymentError.Code.CREATE_PAYMENT_VALIDATION_ERROR;
+
+@Priority(1)
+public class ViolationExceptionMapper implements ExceptionMapper<JerseyViolationException> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ViolationExceptionMapper.class);
+
+    @Override
+    public Response toResponse(JerseyViolationException exception) {
+        LOGGER.error(exception.getMessage());
+        ConstraintViolation<?> firstException = exception.getConstraintViolations().iterator().next();
+        String message = firstException.getMessage();
+        String fieldName = getApiFieldName(firstException);
+        PaymentError paymentError = PaymentError.aPaymentError(fieldName, CREATE_PAYMENT_VALIDATION_ERROR, message);
+
+        return Response.status(422)
+                .entity(paymentError)
+                .build();
+    }
+
+    private String getApiFieldName(ConstraintViolation<?> firstException) {
+        Field field = getField(firstException);
+        return getJacksonPropertyName(field).orElse(field.getName());
+    }
+
+    private Optional<String> getJacksonPropertyName(Field field) {
+        return Arrays.stream(field.getAnnotations())
+                .filter(annotation -> annotation instanceof JsonProperty)
+                .findFirst()
+                .map(annotation -> ((JsonProperty) annotation).value());
+    }
+    
+    private Field getField(ConstraintViolation<?> firstException) {
+        Class<?> leafBean = firstException.getLeafBean().getClass();
+        String fieldName = getFieldNameFromPath(firstException.getPropertyPath());
+        try {
+            return leafBean.getField(fieldName);
+        } catch (NoSuchFieldException e) {
+            LOGGER.error(String.format("Cannot process violation exception. " +
+                    "Field %s does not exist or is not public on class %s", fieldName, leafBean.toString()));
+            throw new WebApplicationException(e);
+        }
+    }
+    
+    private String getFieldNameFromPath(Path path) {
+        String[] pathParts = path.toString().split("\\.");
+        return pathParts[pathParts.length - 1];
+    }
+}
+
--- a/src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
@@ -6,8 +6,13 @@ import io.swagger.annotations.ApiModelProperty;
 import org.apache.commons.lang3.StringUtils;
 import uk.gov.pay.commons.model.charge.ExternalMetadata;
 
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.Size;
 import java.util.Optional;
 import java.util.StringJoiner;
+
+import static uk.gov.pay.api.validation.PaymentRequestValidator.AGREEMENT_ID_MAX_LENGTH;
 
 @ApiModel(value = "CreatePaymentRequest", description = "The Payment Request Payload")
 public class CreatePaymentRequest {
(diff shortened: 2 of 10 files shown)


src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+    
+    private String getLastElementOfPath(Path path) {
+        String[] pathParts = path.toString().split("\\.");
+        return pathParts[pathParts.length - 1];
+    }
+
+    private ConstraintViolation<?> getFirstException(ConstraintViolationException exception) {
+        return exception.getConstraintViolations().iterator().next();
```
oswaldquek: do you really need a method for this line?

oswaldquek: do you really need a method for this line?

src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
```diff
@@ -0,0 +1,71 @@
...
+    private Optional<String> getJacksonPropertyName(Field field) {
+        return Arrays.stream(field.getAnnotations())
+                .filter(annotation -> annotation instanceof JsonProperty)
+                .findFirst()
+                .map(annotation -> ((JsonProperty) annotation).value());
+    }
+    
+    private Field getField(ConstraintViolation<?> firstException) {
+        Class<?> leafBean = firstException.getLeafBean().getClass();
+        String fieldName = getLastElementOfPath(firstException.getPropertyPath());
```
oswaldquek: Purely down to my unfamiliarity with the javax.validation API. Why is the last element of the path got here and not the first?

src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
```diff
@@ -0,0 +1,71 @@
...
+        PaymentError paymentError = PaymentError.aPaymentError(fieldName, CREATE_PAYMENT_VALIDATION_ERROR, message);
+
+        return Response.status(422)
+                .entity(paymentError)
+                .build();
+    }
+
+    private String getApiFieldName(ConstraintViolation<?> firstException) {
+        Field field = getField(firstException);
+        return getJacksonPropertyName(field).orElse(field.getName());
```
oswaldquek: Why can't we simply return the `field.getName`; why do we have to get the jackson property name?

src/main/java/uk/gov/pay/api/model/CreatePaymentRequest.java
```diff
@@ -29,12 +38,23 @@
     public static final String PREFILLED_ADDRESS_CITY_FIELD_NAME = "city";
     public static final String PREFILLED_ADDRESS_POSTCODE_FIELD_NAME = "postcode";
     public static final String PREFILLED_ADDRESS_COUNTRY_FIELD_NAME = "country";
+    
+    @Min(value = AMOUNT_MIN_VALUE, message = "Must be greater than or equal to {value}")
```
oswaldquek: Shouldn't `AMOUNT_MIN_VALUE` be defined in this class (CreatePaymentRequest) rather than PaymentRequestValidator? My understanding is that PaymentRequestValidator will be deprecated.

oswaldquek: Same for AMOUNT_MAX_VALUE, REFERENCE_MAX_LENGTH, DESCRIPTION_MAX_LENGTH and AGREEMENT_ID_MAX_LENGTH.

src/test/java/uk/gov/pay/api/it/validation/PaymentsResourceAmountValidationITest.java
```diff
@@ -48,6 +48,29 @@ public void createPayment_responseWith422_whenAmountIsNegative() throws IOExcept
...
 
+    @Test
+    public void createPayment_responseWith422_whenAmountIsLessThanMin() throws IOException {
+
+        String payload = "{" +
+                "  \"amount\" : 0," +
+                "  \"reference\" : \"Some reference\"," +
+                "  \"description\" : \"Some description\"," +
+                "  \"return_url\" : \"https://somewhere.gov.uk/rainbow/1\"" +
+                "}";
```
oswaldquek: Personally I'd prefer something like https://github.com/alphagov/pay-connector/blob/037141a1da79e72eb365ea950328668eaed7497a/src/test/java/uk/gov/pay/connector/it/resources/CreateGatewayAccountResourceTest.java#L49-L54 but the way it's done currently (by writing out the json explicitly) is still readable for now.

src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
danworth: That's because in the error message we want to use the name used in the api/json request and not just the name of the field on the class, e.g. returning `return_url` and not `returnUrl`. This will look for `@JsonProperty` e.g.  `@JsonProperty("return_url")` on the field which has failed validation and use that if it exists.

src/test/java/uk/gov/pay/api/it/validation/PaymentsResourceAmountValidationITest.java
danworth: Happy to make improvements but will keep them in separate PRs to avoid too much change within the tests we're relying on as we go. I'll look at this again when I come to clean up at the end if you don't mind.

src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
danworth: The last element is the field name, the path will look something like `CreatePaymentRequest.amount` or `CreatePaymentRequest.returnUrl`. Since it isn't clear, perhaps I'll rename this `getFieldNameFromPath`?

oswaldquek: `getFieldNameFromPath` would be clearer. You'd still be doing `pathParts[pathParts.length - 1]` to get this though; I wouldn't be averse to writing a short comment to make it clear why we're doing this. Unfortunately this is a side effect of using reflection....anything goes!

src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
oswaldquek: 👍

src/test/java/uk/gov/pay/api/it/validation/PaymentsResourceAmountValidationITest.java
oswaldquek: Yeah, no need to make the improvement now. I think most of these validation tests can be collapsed but it should be done separately from this PR.

src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
danworth: Trying out ideas from reading 'Clean Code' which promotes the idea of small methods with everything in a method being of a similar logical level. Since the contents of `getFirstException` deals with `iterator().next()` its a level much lower than the other code within `toResponse()` and I believe it keeps reading `toResponse` cleaner, so I think it makes sense to put in it's own method. Unless you feel strongly I'd like to keep it but welcome the discussion.

oswaldquek: I do agree with that idea, especially when the method names describe what is being done, e.g. `getField` and `getJacksonPropertyName` but my view is that if something is perfectly readable, extracting into methods can lead to unreadability. `ConstraintViolation<?> firstException = exception.getConstraintViolations().iterator().next();` isn't hard to understand at all. It's like saying `String s = list.get(0)` can be extracted to `String s = getFirstElement(list)`...

danworth: Ok I've removed the method. I still like the idea of keeping the contents of methods at the same conceptual level and even though something maybe simple to understand, as it is in this case, it can still add to the clutter and cognitive load when first understanding what a method is doing. Using even small single line methods keeps the lower level implementation available for any reader who wishes to know more. Maybe its a case of when we're reviewing we need to know exactly how something works and so want to see the implementation immediately and not dig deeper for it, whilst a later reader might be happy to know the flow of a given method and go deeper as and when needed.

I entirely agree with your point about needlessly heavily nested methods diminishing readability and I've felt this frustration myself, so on balance and since you feel strongly I have removed it, thanks for the discussion, it certainly helps me be more considerate of how people read my code and hopefully strike a better balance 👍

PR #476, opened 2019-05-14
BAU: Don't log at error level for not found errors

heathd:
Not found errors are user errors in the use of the API. They should not be
logged at ERROR level (and don't really need to be logged at all)

## Reviewers

please merge on approval


heathd:
I don't think we need to log warnings either. The access logs will show what happened. 404s are generally not an 'exceptional' case they are incorrect user input.

rhowe-gds:
404s will be logged in the access logs for the service. Provided nothing untoward is detected by the application I wouldn't expect it to log anything at a level higher than `INFO`

heathd:
**REVIEWERS** please merge on approval

PR #479, opened 2019-05-15
Separate integration tests from unit tests

rhowe-gds:
Integration tests are slow. They involve starting a PostgreSQL container and
instantiating Dropwizard.

Split them from the unit tests and run the unit tests with the surefire plugin
instead.

Rename integration tests to match the default include/exclude rules for the
surefire and failsafe maven plugins. By default, the failsafe plugin will
include any tests matching IT*.java, *IT.java and *ITCase.java. This was not
the naming scheme our integration tests had, so fix that.

Include Pact tests in the integration test suite as they spin up a Dropwizard
instance.

Since unit tests are so quick, we no longer need to skip them when running
local builds.

maven's 'package' target doesn't include running integration tests, so switch
to using the 'verify' target during CI builds, which is what we should've been
doing anyway.

### DIFF
--- a/build-local.sh
+++ b/build-local.sh
@@ -4,5 +4,5 @@ set -e
 
 cd "$(dirname "$0")"
 
-mvn -DskipTests clean package
+mvn -DskipITs clean verify
 docker build -t govukpay/publicapi:local .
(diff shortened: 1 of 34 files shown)


build-local.sh
```diff
@@ -4,5 +4,5 @@ set -e
 
 cd "$(dirname "$0")"
 
-mvn -DskipTests clean package
+mvn -DskipITs clean verify
```
kbottla: Does it mean we will now have unit tests run as part of local build ?

oswaldquek: yes

PR #483, opened 2019-05-16
Pp 5100 prefilled cardholder

oswaldquek:
The `POST /v1/payments` endpoint on PaymentsResource now fully uses the javax validation framework to do validations.


src/main/java/uk/gov/pay/api/validation/ExactLengthValidator.java
```diff
@@ -1,10 +1,19 @@
 package uk.gov.pay.api.validation;
 
 
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
-class ExactLengthValidator {
+class ExactLengthValidator implements ConstraintValidator<ValidCountry, String>  {
```
danworth: This was a little confusing at first but I see you've re-used an existing `ExactLengthValidator` class and added the implements `ConstraintValidator...`. To make this reusable perhaps create an annotation `@ExactLength` which is validated by this class and then use it within the `@ValidCountry` or you could just use `@Size` in a similar manner and use `@ReportAsSingleViolation`.
```
@Target({ FIELD, METHOD, PARAMETER, ANNOTATION_TYPE, TYPE_USE })
@Retention(RUNTIME)
@Constraint(validatedBy = {})
@Documented
@Size(max = 2, min = 2)
@ReportAsSingleViolation
public @interface ValidCountry {

    String message() default "Must be exactly 2 characters length";

    Class<?>[] groups() default {};

    Class<? extends Payload>[] payload() default {};
}
```

src/main/java/uk/gov/pay/api/validation/ValidCountry.java
```diff
@@ -0,0 +1,27 @@
...
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target({ FIELD, METHOD, PARAMETER, ANNOTATION_TYPE, TYPE_USE })
+@Retention(RUNTIME)
+@Constraint(validatedBy = ExactLengthValidator.class)
```
danworth: I've added some comments above, I think it would be better to use `@Size` and `@ReportAsSingleViolation` since its easier to extend in future (such as adding other validations like with did with `@ValidMetadata`). I'm not sure its a good idea to have this annotation validated by a class which is very specific which doesn't have a common name, it was a little confusing at first.

src/main/java/uk/gov/pay/api/validation/PaymentRequestValidator.java
```diff
@@ -1,74 +0,0 @@
-package uk.gov.pay.api.validation;
```
danworth: Brilliant, thats the end of the `PaymentRequestValidator`!! 👍

src/main/java/uk/gov/pay/api/validation/ValidCountry.java
oswaldquek: So I used `@Size` originally but there was a test about an empty country being valid (https://github.com/alphagov/pay-publicapi/blob/master/src/test/java/uk/gov/pay/api/validation/PaymentRequestValidatorTest.java#L96) and I couldn't use two `@Size`es, one specifying a char length of 2 and one specifying a char length of 0. This is the reason I reused the existing ExactLengthValidator, which should arguably be renamed CountryLengthValidator.

oswaldquek: There's a sneaky `isBlank(value)` in the ExactLengthValidator that makes an empty country string possible.

src/main/java/uk/gov/pay/api/validation/ExactLengthValidator.java
```diff
@@ -1,10 +1,19 @@
...
+    
     static boolean isValid(String value, int length) {
         return isBlank(value) || value.length() == length;
     }
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        return isValid(value, 2);
```
stephencdaly: It strikes me as odd to default to a length of 2 for this seemingly generic validator.
Also I'm not really sure this or the `ValidCountry` validators are necessary. Why not just add the `@Size(min = 2, max = 2, message = "Must be exactly 2 characters length")` annotation on the country field?

stephencdaly: Oh Dan beat me to it

oswaldquek: We can't use `@Size(min = 2, max = 2, message = "Must be exactly 2 characters length")` because an empty country needs to be valid as well :(
I checked with Sandor who implemented this originally and it's a requirement by Katie.

src/main/java/uk/gov/pay/api/validation/ValidCountry.java
stephencdaly: If it's renamed to CountryValidator, I think it seems fine in that case

src/main/java/uk/gov/pay/api/validation/ExactLengthOrEmpty.java
```diff
@@ -0,0 +1,29 @@
...
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target({ FIELD, METHOD, PARAMETER, ANNOTATION_TYPE, TYPE_USE })
+@Retention(RUNTIME)
+@Constraint(validatedBy = ExactLengthOrEmptyValidator.class)
+@Documented
+public @interface ExactLengthOrEmpty {
+
+    String message() default "Must be exactly 2 characters length";
```
stephencdaly: The message should use the `size`

src/main/java/uk/gov/pay/api/validation/ExactLengthOrEmpty.java
```diff
@@ -0,0 +1,29 @@
...
+
+    Class<?>[] groups() default {};
+
+    Class<? extends Payload>[] payload() default {};
+
+    int size();
```
danworth: Needs rebasing but just a quick comment, but might be nice to call this `length` so its consistent with the wording of the annotation and the attribute in `ExactLengthOrEmptyValidator` which is `length`. Not a bit problem so don't go out of your way to do it if.

oswaldquek: i've fixed this

PR #484, opened 2019-05-16
PP-5133 Add pact test for zero amount not allowed error

stephencdaly:
Test the contract with connector that connector will give the expected error response when a request is made with a 0 amount and the flag on the gateway account is not enabled to allow this.

### DIFF
--- a/src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
+++ b/src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
@@ -11,6 +11,7 @@ import uk.gov.pay.api.app.RestClientFactory;
 import uk.gov.pay.api.app.config.PublicApiConfig;
 import uk.gov.pay.api.app.config.RestClientConfig;
 import uk.gov.pay.api.auth.Account;
+import uk.gov.pay.api.exception.CreateChargeException;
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/api/service/CreatePaymentServiceTest.java
```diff
@@ -216,4 +219,24 @@ public void testCreatePaymentWithPrefilledCardholderDetails() {
         assertThat(billingAddress.getPostcode(), is("AB1 CD2"));
         assertThat(billingAddress.getCountry(), is("GB"));
     }
+    
+    @Test
+    @PactVerification({"connector"})
+    @Pacts(pacts = {"publicapi-connector-create-payment-with-zero-amount-not-allowed"})
+    public void testCreatePaymentWithZeroAmountWhenZeroAmountNotAllowedForAccount() {
+        Account account = new Account("123456", TokenPaymentType.CARD);
+        ValidCreatePaymentRequest requestPayload = new ValidCreatePaymentRequest(CreatePaymentRequest.builder()
```
danworth: Just a heads up that @oswaldquek is likely to be removing `ValidCreatePaymentRequest` and just `CreatePaymentRequest` will remain but it'll be easy to update this test when that time comes.

PR #486, opened 2019-05-16
PP-5199 Use the Jedis client to flush redis

stephencdaly:
Instead of jumping into the container to run `redis-cli`, flush the database
from Java. This should be more efficient and make it easier to notice any
errors which might get thrown.

### DIFF
--- a/src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
+++ b/src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
@@ -128,12 +128,11 @@ class RedisContainer {
     }
 
     public void clearRedisCache() {
-        try {
-            String[] command = {"redis-cli", "FLUSHALL"};
-            String id = docker.execCreate(containerId, command).id();
-            docker.execStart(id);
-        } catch (DockerException | InterruptedException e) {
-            throw new RuntimeException(e);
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/api/it/rule/RedisContainer.java
```diff
@@ -128,12 +128,8 @@ void stop() {
...
     public void clearRedisCache() {
-        try {
-            String[] command = {"redis-cli", "FLUSHALL"};
-            String id = docker.execCreate(containerId, command).id();
-            docker.execStart(id);
-        } catch (DockerException | InterruptedException e) {
-            throw new RuntimeException(e);
-        }
+        Jedis jedis = new Jedis(host, port);
+        jedis.flushAll();
```
rhowe-gds: public String flushAll()
Delete all the keys of all the existing databases, not just the currently selected one. This command never fails.
Specified by:
flushAll in interface BasicCommands
Returns:
Status code reply

rhowe-gds: we should probably do some error checking here

stephencdaly: the return value of `jedis.flushAll()` is `"OK"`...

PR #492, opened 2019-05-20
PP-5100 Remove ValidCreatePaymentRequest

danworth:
- Since validation logic has been added to `CreatePaymentRequest` it is no longer
necessary to have a `ValidCreatePaymentRequest`. This commit removes it.


src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -255,10 +254,7 @@ public Response createNewPayment(@ApiParam(value = "accountId", hidden = true) @
                     "Must be greater than or equal to 1"));
         }
         
-        ValidCreatePaymentRequest validCreatePaymentRequest = new ValidCreatePaymentRequest(createPaymentRequest);
-        logger.info("Payment create request passed validation and parsed to {}", validCreatePaymentRequest);
```
alexbishop1: It’s OK this is removed because we log something similar a few lines above

PR #494, opened 2019-05-21
PP-5224 Split up CreatePaymentRequest

danworth:
CreatePaymentRequest is currently used to represent both Card and DirectDebit payment
requests. This is becoming confusing and impractical as validation requirements
diverge between the two. This commit create dedicated sub-type for each.
1. Create CreatCardPaymentRequest and CreateDirectDebitRequest classes. Each
implementing the attributes and validation require beyond the base class of the
existing CreatePaymentRequest.
2. Simplify ViolationExceptionMapper to convert to the api field name using a
camel case to snake case conversion. This seems safe as it follows our convention
and avoids the complexity of reflection which would need extending to accomadate
the newly introduced hierarchy (if the field is not found on the sub-class then
it would have to search the super class).
3. Remove validation from PaymentResource method and into the new CreateDirect-
DebitPaymentRequest.
4. Reinstate ignored test now that a request can have either return_url or
agreement_id and validate independently.

### DIFF
--- a/src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
@@ -1,6 +1,6 @@
 package uk.gov.pay.api.exception.mapper;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.CaseFormat;
 import io.dropwizard.jersey.validation.JerseyViolationException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -9,12 +9,8 @@ import uk.gov.pay.api.model.PaymentError;
 import javax.annotation.Priority;
 import javax.validation.ConstraintViolation;
 import javax.validation.Path;
-import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
-import java.lang.reflect.Field;
-import java.util.Arrays;
-import java.util.Optional;
 
 import static uk.gov.pay.api.model.PaymentError.Code.CREATE_PAYMENT_VALIDATION_ERROR;
 
@@ -27,7 +23,7 @@ public class ViolationExceptionMapper implements ExceptionMapper<JerseyViolation
         LOGGER.error(exception.getMessage());
         ConstraintViolation<?> firstException = exception.getConstraintViolations().iterator().next();
         String message = firstException.getMessage();
-        String fieldName = getApiFieldName(firstException);
+        String fieldName = getApiFieldName(firstException.getPropertyPath());
         PaymentError paymentError = PaymentError.aPaymentError(fieldName, CREATE_PAYMENT_VALIDATION_ERROR, message);
 
         return Response.status(422)
@@ -35,30 +31,8 @@ public class ViolationExceptionMapper implements ExceptionMapper<JerseyViolation
                 .build();
     }
 
-    private String getApiFieldName(ConstraintViolation<?> firstException) {
-        Field field = getField(firstException);
-        return getJacksonPropertyName(field).orElse(field.getName());
-    }
-
-    private Optional<String> getJacksonPropertyName(Field field) {
-        return Arrays.stream(field.getAnnotations())
-                .filter(annotation -> annotation instanceof JsonProperty)
-                .findFirst()
-                .map(annotation -> ((JsonProperty) annotation).value());
-    }
-    
-    private Field getField(ConstraintViolation<?> firstException) {
-        Class<?> leafBean = firstException.getLeafBean().getClass();
-        String fieldName = getFieldNameFromPath(firstException.getPropertyPath());
-        try {
-            Field f = leafBean.getDeclaredField(fieldName);
-            f.setAccessible(true);
-            return f;
-        } catch (NoSuchFieldException e) {
-            LOGGER.error(String.format("Cannot process violation exception. " +
-                    "Field %s does not exist or is not public on class %s", fieldName, leafBean.toString()));
-            throw new WebApplicationException(e);
-        }
+    private String getApiFieldName(Path path) {
+        return CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, getFieldNameFromPath(path));
     }
     
     private String getFieldNameFromPath(Path path) {
--- a/src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
@@ -1,0 +1,121 @@
+package uk.gov.pay.api.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+import uk.gov.pay.api.utils.JsonStringBuilder;
+import uk.gov.pay.api.validation.ValidReturnUrl;
+import uk.gov.pay.commons.model.charge.ExternalMetadata;
+
+import javax.validation.Valid;
+import javax.validation.constraints.Size;
+import java.util.Optional;
+import java.util.StringJoiner;
+
+@ApiModel(value = "CreatePaymentRequest", description = "The Payment Request Payload")
+public class CreateCardPaymentRequest extends CreatePaymentRequest {
+
+    public static final String RETURN_URL_FIELD_NAME = "return_url";
+    public static final int URL_MAX_LENGTH = 2000;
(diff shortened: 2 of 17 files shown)


src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
```diff
@@ -12,9 +13,11 @@
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
+import java.lang.reflect.AccessibleObject;
```
alexbishop1: Unused imports?

src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
```diff
@@ -27,38 +30,16 @@ public Response toResponse(JerseyViolationException exception) {
...
-            Field f = leafBean.getDeclaredField(fieldName);
-            f.setAccessible(true);
-            return f;
-        } catch (NoSuchFieldException e) {
-            LOGGER.error(String.format("Cannot process violation exception. " +
-                    "Field %s does not exist or is not public on class %s", fieldName, leafBean.toString()));
-            throw new WebApplicationException(e);
-        }
+    private String getApiFieldName(Path path) {
+        return CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, getFieldNameFromPath(path));
```
alexbishop1: 🐪

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
```diff
@@ -0,0 +1,125 @@
...
+        this.metadata = createPaymentRequestBuilder.getMetadata();
+    }
+
+    @Override
+    public TokenPaymentType getRequestType() {
+        return TokenPaymentType.CARD;
+    }
+
+    @ApiModelProperty(value = "service return url", required = true, example = "https://service-name.gov.uk/transactions/12345")
+    public Optional<String> getReturnUrl() {
```
alexbishop1: A create card payment request should always have a return URL, so I don’t think this should return an optional

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
```diff
@@ -0,0 +1,125 @@
...
+        return Optional.ofNullable(returnUrl);
+    }
+
+    public Optional<String> getAgreementId() {
```
alexbishop1: Does this need to be here at all?

src/main/java/uk/gov/pay/api/model/CreateDirectDebitPaymentRequest.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    @Min(value = 1, message = "Must be greater than or equal to 1")
+    public int getAmount() {
+        return super.getAmount();
+    }
+
+    @ApiModelProperty(value = "ID of the agreement being used to collect the payment", required = false, example = "33890b55-b9ea-4e2f-90fd-77ae0e9009e2")
+    public Optional<String> getAgreementId() {
```
alexbishop1: An create on-demand Direct Debit payment request should always have an agreement ID

src/main/java/uk/gov/pay/api/model/CreatePaymentRequestBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+    private String cardholderName;
+    private String addressLine1;
+    private String addressLine2;
+    private String city;
+    private String postcode;
+    private String country;
+    private PrefilledCardholderDetails prefilledCardholderDetails;
+
+    public CreatePaymentRequest build() {
+        return this.getAgreementId() == null ? new CreateCardPaymentRequest(this) : new CreateDirectDebitPaymentRequest(this);
```
alexbishop1: It feels weird to pass the builder to these constructors. It would be more conventional to have the builder extract the relevant fields from itself and then pass them to a multi-argument constructor. Alternatively, the builder could be renamed to something else (I suppose it’s a data source of some sort).

src/main/java/uk/gov/pay/api/model/CreateDirectDebitPaymentRequest.java
```diff
@@ -0,0 +1,71 @@
...
+    public CreateDirectDebitPaymentRequest(CreatePaymentRequestBuilder createPaymentRequestBuilder) {
+        super(createPaymentRequestBuilder);
+        this.agreementId = createPaymentRequestBuilder.getAgreementId();
+    }
+
+    @Override
+    @Min(value = 1, message = "Must be greater than or equal to 1")
+    public int getAmount() {
+        return super.getAmount();
+    }
```
alexbishop1: Isn’t this essentially identical to what’s in the superclass?

src/main/java/uk/gov/pay/api/service/ConnectorUriGenerator.java
```diff
@@ -24,9 +27,9 @@ public ConnectorUriGenerator(PublicApiConfig configuration) {
         this.configuration = configuration;
     }
 
-    String chargesURI(Account account, String agreementId) {
+    String chargesURI(Account account, CreatePaymentRequest createCardPaymentRequest) {
```
alexbishop1: I think this could just take `TokenPaymentType`

src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
```diff
@@ -431,6 +431,7 @@ public void createPayment_responseWith422_whenZeroAmountForDirectDebitAccount()
                 .add("reference", REFERENCE)
                 .add("description", DESCRIPTION)
                 .add("return_url", RETURN_URL)
+                .add("agreement_id", "1234")
```
alexbishop1: The request should have either a return URL or an agreement ID, not both

src/main/java/uk/gov/pay/api/exception/mapper/ViolationExceptionMapper.java
danworth: 👍

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
danworth: Yeah thought about that on the train home last night, same for agreementId in the other one, 👍

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
danworth: No, left over from when it was abstract on the base class.

src/main/java/uk/gov/pay/api/model/CreateDirectDebitPaymentRequest.java
danworth: yep 👍

src/main/java/uk/gov/pay/api/model/CreateDirectDebitPaymentRequest.java
danworth: No, the min amount on the superclass is 0, for DirectDebit we want it to be 1. This overrides the validation to 1.

src/main/java/uk/gov/pay/api/service/ConnectorUriGenerator.java
danworth: yes good idea.

src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
danworth: ah yes 👍

src/main/java/uk/gov/pay/api/model/CreatePaymentRequestBuilder.java
danworth: I'm not keen on deconstructing the builder to pass to the constructors as it'll have too many arguments and be confusing - we'd also need a few overloaded constructors since some attributes are optional and then it just becomes a mess. I feel that passing a builder to a constructor is good practice and the builder _is_ a builder so I'm not entirely sure about renaming it.

danworth:
> This looks a lot clearer. Left a few comments.
> 
> Is it currently the case that a request with both a `"return_url"` and an `"agreement_id"` is rejected? Does this still hold after these changes are applied?

Currently (master) the parser will look for an `agreement_id` and if present it will set the `return_url` to null, whether or not it was actually present in the request. We maintain this behaviour since if the `agreement_id` is present it creates a `CreateDirectDebitPaymentRequest` which has no `return_url` or a `CreateCardPaymentRequest` which has no `agreement_id`. We don't in this PR, nor did we previously, return an error if both `return_url` and `agreement_id` are present.

PR #497, opened 2019-05-28
PP-5242 Add pact test for collect DD payment

stephencdaly:
- There is currently no pact test for collecting from an existing direct debit mandate.
- There is a pact test for creating a one-off DD payment, but we are dropping support for one-off so this should be removed.
- Replace pact test to create a one-off DD payment with a pact test to collect a payment for an existing mandate.


src/test/resources/pacts/publicapi-direct-debit-connector-collect-payment.json
```diff
@@ -43,30 +43,12 @@
             "status": "created",
             "finished": false
           },
-          "created_date": "2016-01-01T12:00:00Z",
+          "created_date": "2010-12-31T22:59:59.132Z",
```
alexbishop1: A long, long time ago…

PR #510, opened 2019-06-11
PP-5292-Add-links-without-next_url to pact file

nimalank7:
## WHAT YOU DID

This adds the links field to publicapi-connector-search-payment-by-first-digits-card-number.json pact file.

This is because we removed the next_url and next_url post from the JSON which Connector emits to publicapi. This requires a modification of the pact file in publicapi - we have to remove the links field first to ensure that publicapi isn't expecting next_url and next_url_post then we have to add the links field without next_url and next_url_post to ensure that the pact file is properly modified.

### DIFF
--- a/src/test/resources/pacts/publicapi-connector-search-payment-by-first-digits-card-number.json
+++ b/src/test/resources/pacts/publicapi-connector-search-payment-by-first-digits-card-number.json
@@ -48,6 +48,18 @@
               "description": "Test description",
               "reference": "aReference",
               "language": "en",
+              "links": [
+                {
+                  "rel": "self",
+                  "method": "GET",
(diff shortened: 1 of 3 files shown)


src/test/resources/pacts/publicapi-connector-search-payment-by-first-digits-card-number.json
```diff
@@ -48,6 +48,18 @@
               "description": "Test description",
               "reference": "aReference",
               "language": "en",
+              "links": [
```
kbottla: We will also need similar change in
`publicapi-connector-search-payment-by-cardholder-name`
`publicapi-connector-search-payment-by-last-digits-card-number`

nimalank7: Apologies I did a rebase and it lost those changes. I've now updated it now.

PR #531, opened 2019-06-18
PP-5298 Add missing Create Payment Information

RoryMalcolm:
- Will now respond with the new information from DD Connector in
  public api


src/test/java/uk/gov/pay/api/it/CachingAuthenticatorIT.java
```diff
@@ -101,9 +101,11 @@ private String aPayment() {
                 .add("amount", 100)
                 .add("language", "en")
                 .add("reference", "ref 12")
-                .add("state", new PaymentState("created", false, null, null))
+                .add("state", new PaymentState("created", false, null, null, null))
                 .add("email", "test@example.com")
                 .add("description", "description")
+                .add("mandate_id", "amandateid")
+                .add("provider_id", "aproviderid")
```
stephencdaly: This looks like it's just for card payments so don't think we should add these her.e

src/main/java/uk/gov/pay/api/model/PaymentState.java
```diff
@@ -24,28 +24,33 @@
...
+                node.has("details") ? node.get("details").asText() : null
         );
     }
 
     public PaymentState() {
     }
 
     public PaymentState(String status, boolean finished) {
-        this(status, finished, null, null);
+        this(status, finished, null, null, null);
```
stephencdaly: I think maybe it's worthwhile having a subclass for direct debit payments states that add the "details" field to make it clear that this is just for direct debits

PR #535, opened 2019-06-18
Pp 5297 validations for create mandate

oswaldquek:


src/main/java/uk/gov/pay/api/model/directdebit/agreement/CreateMandateRequest.java
```diff
@@ -5,19 +5,29 @@
...
 
+    @NotNull
+    @ValidReturnUrl
+    @Length(max = 2000, message = "Must be less than or equal to {max} characters length")
     @JsonProperty(value = "return_url")
     private String returnUrl;
-    
+
+    @NotNull
+    @Length(min = 1, max = 255, message = "Must have a size between {min} and {max}")
```
stephencdaly: I think `@NotBlank` is better as it will prevent blank strings too. Then you'd only need the a max length

oswaldquek: fixed

src/main/java/uk/gov/pay/api/model/directdebit/agreement/CreateMandateRequest.java
```diff
@@ -5,19 +5,32 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
+import org.hibernate.validator.constraints.Length;
```
danworth: Just a note, it might be better to use `@Size` which is included in the validation spec `javax.validation.constraints` and is implementation agnostic. Hibernate `@Length` is hibernate specific version of `@Size` and it ties us to use hibernate. Bit more info here: https://www.baeldung.com/jpa-size-length-column-differences

oswaldquek: fixed

stephencdaly: can you use Size for the other fields too?

oswaldquek: argh i got confused. i've used Size all over now.

PR #537, opened 2019-06-19
PP-5298 Add validations for create payment request

stephencdaly:
Add validations on the fields for the create direct debit payment request and add tests for these.


src/test/java/uk/gov/pay/api/resources/directdebit/DirectDebitPaymentsResourceIT.java
```diff
@@ -0,0 +1,179 @@
...
+                "mandate_id", MANDATE_ID));
+
+        postPaymentResponse(request)
+                .statusCode(201)
+                .body("$", not(hasKey("description")));
+    }
+
+    @Test
+    @Parameters(method = "parametersForValidation")
+    public void validationFailures(
```
stephencdaly: The test class was renamed from DirectDebitPaymentsResourceTest and this validationFailures test added. Git wasn't clever enough to recognise the rename

src/test/java/uk/gov/pay/api/resources/directdebit/DirectDebitResourceITBase.java
```diff
@@ -0,0 +1,94 @@
...
+
+import static io.dropwizard.testing.ConfigOverride.config;
+import static io.dropwizard.testing.ResourceHelpers.resourceFilePath;
+import static io.restassured.RestAssured.given;
+import static io.restassured.http.ContentType.JSON;
+import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
+import static uk.gov.pay.api.model.TokenPaymentType.DIRECT_DEBIT;
+import static uk.gov.pay.commons.testing.port.PortFactory.findFreePort;
+
+public class DirectDebitResourceITBase {
```
stephencdaly: Extracted this out from DirectDebitPaymentsResourceTest. Can be used by other DD Resource tests

src/test/java/uk/gov/pay/api/resources/directdebit/DirectDebitPaymentsResourceIT.java
```diff
@@ -0,0 +1,181 @@
...
+
+        postPaymentResponse(new Gson().toJson(payload))
+                .statusCode(422)
+                .contentType(JSON)
+                .body("code", is(errorCode))
+                .body("field", is(field))
+                .body("description", is(expectedErrorMessage));
+    }
+
+    private Object[] parametersForValidation() {
```
danworth: Just a passing comment before I go for lunch, happy to carry on when I'm back. I find this a little difficult to read at first. Might it be nice to create an inner class to represent these parameters and then return an array of them, you could even have a `toPayload` method on such class which will simplify the test method itself...e.g `var payload = parameters.toPayload()` and `.body("description", is(parameters.getExpectedMessage())` the test method itself becomes something like 
`public void validationFailures(TestParameters parameters)`

Happy to explain is that doesn't make sense :)

src/test/java/uk/gov/pay/api/resources/directdebit/DirectDebitPaymentsResourceIT.java
```diff
@@ -0,0 +1,289 @@
...
+                        .withErrorCode("P0102")
+                        .build(),
+        };
+    }
+
+    String paymentEventsLocationFor(String chargeId) {
+        return configuration.getBaseUrl() + "v1/directdebit/payments/" + chargeId + "/events";
+    }
+
+    private String paymentLocationFor(String chargeId) {
```
RoryMalcolm: why's this private when the rest aren't?

stephencdaly: hmm should just be in the base class anyway, i'll update

PR #540, opened 2019-06-19
PP-5298 Add further pact tests

RoryMalcolm:
- Now tests to ensure DD payments without descriptions can be created


oswaldquek:
I'm not really sure about adding a pact test for an optional field. Pact tests are rather expensive and if we're testing that description is optional on the dd-connector side via an integration test that should suffice.

RoryMalcolm:
> Pact tests are rather expensive and if we're testing that description is optional on the dd-connector side via an integration test that should suffice.

I think you're right, will close

PR #549, opened 2019-06-21
PP-5299: Return bank statement reference when getting a mandate

oswaldquek:
There are some TODOs in this PR which I intend to address in the next PR.


src/main/java/uk/gov/pay/api/model/directdebit/mandates/GetMandateResponse.java
```diff
@@ -7,74 +7,77 @@
...
+        this.mandateReference = mandateReference;
         this.state = state;
         this.links = links;
     }
 
     public static GetMandateResponse from(MandateConnectorResponse mandate, MandateLinks links) {
         return new GetMandateResponse(
                 mandate.getMandateId(),
-                mandate.getMandateReference(),
+                null,
```
oswaldquek: this is null at the moment as dd-connector doesn't return the provider id. this will be fixed in an upcoming PR.

PR #553, opened 2019-06-21
PP-5299: Return provider id

oswaldquek:
Return provider id when getting a confirmed mandate from dd-connector.


src/main/java/uk/gov/pay/api/model/directdebit/mandates/GetMandateResponse.java
```diff
@@ -35,7 +35,7 @@ private GetMandateResponse(String agreementId,
     public static GetMandateResponse from(MandateConnectorResponse mandate, MandateLinks links) {
         return new GetMandateResponse(
                 mandate.getMandateId(),
-                null,
+                mandate.getProviderId(),
```
stephencdaly: Can a model be shared for the create mandate and get mandate response?

oswaldquek: They can. I'll do a new PR for it.

PR #559, opened 2019-06-24
PP-5299: Refactor to use common base class

oswaldquek:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation


src/main/java/uk/gov/pay/api/model/directdebit/mandates/CreateMandateResponse.java
```diff
@@ -18,13 +18,7 @@
...
 
-    private final String mandateId;
-    private final String providerId;
-    private final String reference;
-    private final String returnUrl;
     private final String createdDate;
-    private final MandateStatus state;
-    private final MandateLinks links;
     private final String description;
     private final String paymentProvider;
```
stephencdaly: Shouldn't the get mandate response have a description, payment provider and created date too?

Same comment as for the dd-connector change, is there an advantage to having 2 models over just having 1 where some fields won't be populated some of the time?

oswaldquek: fixed

oswaldquek:
@stephencdaly soz can you re-approve. tests were failing and i had to rebase

PR #576, opened 2019-07-04
PP-5377 Return 409 On Create Payment With Invalid Mandate State

RoryMalcolm:
- When a user tries to create a payment on a mandate which has an invalid
  state, they should receive a 409 error with a related error code.

### DIFF
--- a/src/test/java/uk/gov/pay/api/exception/mapper/CreateChargeExceptionMapperTest.java
+++ b/src/test/java/uk/gov/pay/api/exception/mapper/CreateChargeExceptionMapperTest.java
@@ -2,6 +2,9 @@ package uk.gov.pay.api.exception.mapper;
 
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
 import uk.gov.pay.api.exception.ConnectorResponseErrorException.ConnectorErrorResponse;
 import uk.gov.pay.api.exception.CreateChargeException;
 import uk.gov.pay.api.model.PaymentError;
@@ -16,7 +19,12 @@ import static org.mockito.Mockito.when;
 import static uk.gov.pay.api.model.PaymentError.Code.CREATE_PAYMENT_MANDATE_STATE_INVALID;
 import static uk.gov.pay.api.model.PaymentError.aPaymentError;
 
+@RunWith(MockitoJUnitRunner.class)
 public class CreateChargeExceptionMapperTest {
+
+
+    @Mock
+    private Response mockResponse = mock(Response.class);
 
     private CreateChargeExceptionMapper mapper;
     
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/api/exception/mapper/CreateChargeExceptionMapperTest.java
```diff
@@ -0,0 +1,44 @@
...
+    private CreateChargeExceptionMapper mapper;
+    
+    @Before
+    public void setUp() {
+        mapper = new CreateChargeExceptionMapper();
+    }
+    
+    @Test
+    public void shouldThrow409_whenMandateStateInvalid(){
+        Response response = mock(Response.class);
```
alexbishop1: Usually we’d put a mock like this in a field with an annotation:

```java
@Mock private Response mockResponse;
```

You’ll need to add a class-level `@RunWith(MockitoJUnitRunner.class)` annotation for this to work.

src/test/java/uk/gov/pay/api/exception/mapper/CreateChargeExceptionMapperTest.java
```diff
@@ -0,0 +1,44 @@
...
+        when(response.readEntity(ConnectorErrorResponse.class))
+                .thenReturn(new ConnectorErrorResponse(
+                        ErrorIdentifier.MANDATE_STATE_INVALID, null, null));
+        
+        Response returnedResponse = mapper.toResponse(new CreateChargeException(response));
+        PaymentError returnedError = (PaymentError) returnedResponse.getEntity();
+        assertThat(returnedResponse.getStatus(), is(409));
+        assertThat(returnedError.getDescription(), 
+                is(aPaymentError(CREATE_PAYMENT_MANDATE_STATE_INVALID).getDescription()));
+        assertThat(returnedError.getCode(), is(aPaymentError(CREATE_PAYMENT_MANDATE_STATE_INVALID).getCode()));
```
alexbishop1: Rather than creating the expected `PaymentError` every time, you could create it once and assign it to a field called `expectedError` or something and reuse that

src/test/java/uk/gov/pay/api/exception/mapper/CreateChargeExceptionMapperTest.java
```diff
@@ -0,0 +1,44 @@
...
+    @Test
+    public void shouldThrow409_whenMandateStateInvalid(){
+        Response response = mock(Response.class);
+        when(response.readEntity(ConnectorErrorResponse.class))
+                .thenReturn(new ConnectorErrorResponse(
+                        ErrorIdentifier.MANDATE_STATE_INVALID, null, null));
+        
+        Response returnedResponse = mapper.toResponse(new CreateChargeException(response));
+        PaymentError returnedError = (PaymentError) returnedResponse.getEntity();
+        assertThat(returnedResponse.getStatus(), is(409));
```
alexbishop1: It’s always nice to have a blank line before the assertions so it’s clear what’s actually being tested. Personally, I like to space things so there’s the stubbing, then a blank line, then the actual call that’s tested, then a blank line, then the assertions and verifications. Like this:

```java
@Test
public void myAmazingTest() {
    when(mockSomething.getSomething()).thenReturn(foo);
    when(mockSomethingElse.getSomethingElse()).thenReturn(bar);

    String result = thisIsTheActualThingWeAreTesting.allOnItsOwnSoItsClear(whatExactlyIsBeingTested);

    assertThat(result, is(whatWeExpect));
    assertThat(result, is(awesome));

    verify(test.isEasyToRead());
    verify(and.weAre(done));
}
```

PR #577, opened 2019-07-05
bau: update java commons library

oswaldquek:
There's a new validations module.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation


src/main/java/uk/gov/pay/api/model/search/directdebit/DirectDebitSearchPaymentsParams.java
```diff
@@ -34,11 +33,9 @@
     private String mandateId;
     
     @QueryParam("from_date")
-    @ValidDate
```
alexbishop1: Why have these annotations gone?

oswaldquek: It was a mistake, I've put them back in now.

PR #580, opened 2019-07-09
PP-5360 get one transaction from ledger

jankowiakmaria:
Initial draft:
* added contract tests for ledger
* initial draft of the changes to the GetPaymentService
* feature flag for ^^
* first round of refactoring


jankowiakmaria:
As a little bit of context - to try to make ledger as close to connector as possible, the contract files are a copy of the contracts for connector.

src/main/java/uk/gov/pay/api/model/TransactionResponse.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    public CardDetails getCardDetails() {
+        return cardDetails;
+    }
+
+    public String getGatewayTransactionId() {
+        return gatewayTransactionId;
+    }
+}
```
mrlumbu: My understanding is that the purpose of introducing TransactionResponse and Charge is to allow you to create an abstraction over the two different apis, so that you get a 
`ChargeFromResponse` from connector
and a
`TransactionResponse` from ledger
and both can be converted into a `Charge`
I'm not sure that we need to do this - it seems the structure of all three of these  classes is basically the same? I may have got the wrong end of the stick though, so let's discuss this.

src/main/java/uk/gov/pay/api/service/GetPaymentService.java
```diff
@@ -1,49 +1,77 @@
...
+                             ConnectorUriGenerator connectorUriGenerator, LedgerUriGenerator ledgerUriGenerator,
+                             PublicApiConfig config) {
         this.client = client;
         this.publicApiUriGenerator = publicApiUriGenerator;
         this.connectorUriGenerator = connectorUriGenerator;
+        this.ledgerUriGenerator = ledgerUriGenerator;
+        this.config = config;
     }
 
     public PaymentWithAllLinks getPayment(Account account, String paymentId) {
```
mrlumbu: It might be nice to move this towards a strategy pattern - especially as we introduce more feature flags etc to manage how we do various operations. But happy with this implementation for now

src/test/java/uk/gov/pay/api/model/PaymentTest.java
```diff
@@ -46,8 +46,9 @@ public void shouldNotPrintEmailAndCardDetailsWhenToString() throws IOException {
                 "    \"finished\": false\n" +
                 "  }\n" +
                 "}", ChargeFromResponse.class);
+        Charge charge = Charge.from(paymentFromConnector);
 
-        PaymentWithAllLinks payment = PaymentWithAllLinks.valueOf(paymentFromConnector, selfUri, eventsUri, cancelUri, refundsUri, captureUri);
+        PaymentWithAllLinks payment = PaymentWithAllLinks.valueOf(charge, selfUri, eventsUri, cancelUri, refundsUri, captureUri);
```
mrlumbu: I have no idea what the payment - charge distinction is in this codebase. Maybe something to think about as you make your changes. If there is no distinction we should settle on one term - probably payment

src/test/java/uk/gov/pay/api/service/GetPaymentServiceLedgerTest.java
```diff
@@ -0,0 +1,174 @@
...
+        PublicApiUriGenerator publicApiUriGenerator = new PublicApiUriGenerator(mockConfiguration);
+        ConnectorUriGenerator connectorUriGenerator = new ConnectorUriGenerator(mockConfiguration);
+        LedgerUriGenerator ledgerUriGenerator = new LedgerUriGenerator(mockConfiguration);
+        Client client = RestClientFactory.buildClient(new RestClientConfig(false));
+        getPaymentService = new GetPaymentService(client, publicApiUriGenerator, connectorUriGenerator, ledgerUriGenerator, mockConfiguration);
+    }
+
+    @Test
+    @PactVerification({"ledger"})
+    @Pacts(pacts = {"publicapi-ledger-get-payment-with-metadata"})
```
mrlumbu: I realise you are duplicating existing patterns, but I think we need to rethink this pattern of creating a new pact for every variant of thing that can be returned from connector. We should probably just have one test for the maximal set of all attributes and leave it at that unless there is a compelling reason not to

src/main/java/uk/gov/pay/api/model/TransactionResponse.java
jankowiakmaria: Your understanding is correct.

At the moment the structure is the same, but I would prefer no to tie ledger API to connector API (it seems to risky).
It's not always easy to say when it's better to have deliberate duplication of the code to make sure we don't mix things up.

src/test/java/uk/gov/pay/api/model/PaymentTest.java
jankowiakmaria: Can change it - just wanted the code to work in the first pass ;)

src/main/java/uk/gov/pay/api/service/GetPaymentService.java
jankowiakmaria: Makes sense 👍 

Just drafted it quickly to have something working.

src/test/java/uk/gov/pay/api/service/GetPaymentServiceLedgerTest.java
jankowiakmaria: yep - was considering min vs max set of attributes, but thought it will be easier to leave as is for the benefit of "connector" compatibility.

jankowiakmaria:
Thanks for review!

PR #582, opened 2019-07-12
Pp 4769 receive payment notifications

nimalank7:
This PR adds an endpoint to publicapi (v1/payment_notifications) to handle telephone payments.

The JSON payload is represented by the CreateTelephonePayment class. It looks like this:

`{
  "amount": 12000,
  "reference": "MRPC12345",
  "description": "New passport application",
  "created_date": "2018-02-21T16:04:25Z",
  "authorised_date": "2018-02-21T16:05:33Z",
  "processor_id": "183f2j8923j8",
  "provider_id": "17498-8412u9-1273891239",
  "auth_code": "666",
  "payment_outcome": {
    "status": "success",
    "code": "P0010",
    "supplemental": {
      "error_code": "ECKOH01234",
      "error_message": "textual message describing error code"
    }
  },
  "card_type": "master-card",
  "name_on_card": "Jane Doe",
  "email_address": "jane_doe@example.com",
  "card_expiry": "02/19",
  "last_four_digits": "1234",
  "first_six_digits": "654321",
  "telephone_number": "+447700900796"
}`

When the JSON payload is successfully validated it returns **only** a 201 response. When validation is unsuccessful it returns a 422 with a response body provided by PaymentError. The required fields are:

- Amount
- Reference
- Description
- Processor id
- Provider id
- Card Type
- Card Expiry
- Last Four Digits
- First Six Digits

Amount has both minimum and maximum validation matching that of CreateCardPaymentRequest. Both reference and description have a maximum character length of 255. Both Processor id and provider id have a @NotNull annotation. (There is an edge case where the user submits the wrong type - we will have to use a custom Jackson deserializer to handle this specifically in a later story).

There are custom validators defined for Card Type, Last Four Digits and First Six Digits. (Note for card expiry I haven't enforced the requirement that it must be later than today's date).

## How to test

I have written a representation test for the deserialization of the JSON payload in CreateTelephonePaymentRequestTest. I have also written integration tests for checking that the custom validation is working correctly - see CreateTelephonePaymentCardExpiryValidation, CreateTelephonePaymentFirstSixCardDigitsValidation and CreateTelephonePaymentLastFourCardDigitsValidation in it/validation/telephone. I have also added an integration test for returning 201 - see GetTelephonePaymentIT in it/telephone.


src/main/java/uk/gov/pay/api/model/telephone/CreateTelephonePaymentRequest.java
```diff
@@ -0,0 +1,162 @@
...
+
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+public class CreateTelephonePaymentRequest {
+
+    public static final int REFERENCE_MAX_LENGTH = 255;
+    public static final int AMOUNT_MAX_VALUE = 10000000;
```
mrlumbu: I'm a bit uneasy about how we're just scattering this round our codebase as a hardcoded value, although I realise you are following precedent here. Just consolidating this into a single static class would be a good start, maybe in a follow up PR, or alternatively create an Amount type that can be validated and use it everywhere

src/main/java/uk/gov/pay/api/resources/telephone/TelephonePaymentNotificationResource.java
```diff
@@ -0,0 +1,34 @@
...
+@Api(value = "/", description = "Public Api Endpoints to receive Telephone Payments")
+public class TelephonePaymentNotificationResource {
+
+    private static final Logger logger = LoggerFactory.getLogger(TelephonePaymentNotificationResource.class);
+    
+    @POST
+    @Timed
+    @Path("/v1/payment_notification")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
```
mrlumbu: presumably we'll need the same swagger annotations we have on PaymentsResource on this one - can be done later

src/main/java/uk/gov/pay/api/validation/CardExpiryValidator.java
```diff
@@ -0,0 +1,21 @@
...
+    private Pattern pattern = Pattern.compile("(0[1-9]|1[0-2])\\/\\d{2}");
+    
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        
+        if (value == null) {
+            return false;
+        }
+        
+        return pattern.matcher(value).matches() ? true : false;
```
mrlumbu: can just be `return pattern.matcher(value).matches();`

src/main/java/uk/gov/pay/api/validation/CardLastFourDigitsValidator.java
```diff
@@ -0,0 +1,21 @@
...
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        
+        if (value == null) {
+            return false;
+        }
+
+        return pattern.matcher(value).matches() ? true : false;
+
+    }
+}
```
mrlumbu: You could have the three pattern based validators extend a common `StringPatternValidator` which would just allow you to set the pattern to be validated, to make this a bit DRYer, and/or a DigitSequenceValidator that takes the number of digits to be validated (not actually sure how to do this off top of my head, but worth exploring)

src/test/java/uk/gov/pay/api/it/validation/telephone/CreateTelepehonePaymentLastFourCardDigitsValidation.java
```diff
@@ -0,0 +1,66 @@
...
+                "  \"processor_id\" : \"1PROC\"," +
+                "  \"provider_id\" : \"1PROV\"," +
+                "  \"card_type\" : \"visa\"," +
+                "  \"card_expiry\" : \"01/99\"," +
+                "  \"last_four_digits\" : null," +
+                "  \"first_six_digits\" : \"123456\"" +
+                "}";
+
+        postPaymentResponse(payload).statusCode(422);
+    }
```
mrlumbu: I think all these tests can ResourceTests
https://www.dropwizard.io/1.3.11/docs/manual/testing.html
This would mean you wouldn't actually have to so spin up the app, so the tests would be much faster.

swagger/swagger.json
```diff
@@ -1332,6 +1366,23 @@
...
+          "minLength" : 0,
+          "maxLength" : 255
+        },
+        "billing_address" : {
+          "description" : "prefilled billing address",
+          "readOnly" : true,
+          "$ref" : "#/definitions/Address"
+        }
+      }
+    },
```
mrlumbu: I'm confused why some of these things are being added as part of this PR?

src/main/java/uk/gov/pay/api/validation/CardExpiryValidator.java
nimalank7: Done!

src/main/java/uk/gov/pay/api/model/telephone/CreateTelephonePaymentRequest.java
nimalank7: I agree - I'll do this in a follow up PR.

src/main/java/uk/gov/pay/api/resources/telephone/TelephonePaymentNotificationResource.java
heathd: I suggest we leave this undocumented for now

nimalank7: Great!

src/main/java/uk/gov/pay/api/validation/CardLastFourDigitsValidator.java
nimalank7: I'm going to leave this for now as I'm happy as it is (and also getting the types to line up was a real pain and it gains little).

src/test/java/uk/gov/pay/api/it/validation/telephone/CreateTelepehonePaymentLastFourCardDigitsValidation.java
nimalank7: These have been added

swagger/swagger.json
nimalank7: For some odd reason the swagger file changed when I made my code - I must of accidentally committed all. At any rate I've rebased and the swagger file has been changed since then.

jankowiakmaria:
Maybe worth adding tests for `400`.

src/test/java/uk/gov/pay/api/resources/telephone/ValidationResourceTest.java
```diff
@@ -0,0 +1,69 @@
...
+                .builder()
+                .addResource(new AuthDynamicFeature(new OAuthCredentialAuthFilter.Builder<Account>()
+                        .setAuthenticator(new AccountAuthenticator(client, mockPublicApiConfig))
+                        .setPrefix("Bearer")
+                        .buildAuthFilter()))
+                .addProvider(new AuthValueFactoryProvider.Binder<>(Account.class))
+                .addProvider(ViolationExceptionMapper.class)
+                .addResource(new TelephonePaymentNotificationResource())
+                .build();
+    }
```
mrlumbu: nice one! Not as simple as I thought it would be 😬

src/test/java/uk/gov/pay/api/resources/telephone/TelephonePaymentResourceITBase.java
```diff
@@ -0,0 +1,95 @@
+package uk.gov.pay.api.resources.telephone;
```
mrlumbu: should probably move this to sit with the test that extends it, but feel free to do this in a future commit

src/main/java/uk/gov/pay/api/resources/telephone/TelephonePaymentNotificationResource.java
```diff
@@ -0,0 +1,38 @@
...
+@Path("/")
+@Api(value = "/", description = "Public Api Endpoints to receive Telephone Payments")
+public class TelephonePaymentNotificationResource {
+
+    private static final Logger logger = LoggerFactory.getLogger(TelephonePaymentNotificationResource.class);
+    
+    public TelephonePaymentNotificationResource(){
+        
+    }
+
```
mrlumbu: logger and empty constructor are unused and can be removed (but feel free to do this in follow up PR)

PR #601, opened 2019-07-23
PP-5453 Update Direct Debit tests to check for state.details

alexbishop1:
Update tests that check what public API returns in response for requests for mandates and payments (both single and search results) to check for the `state.details` property, which is passed through
from Direct Debit connector.


src/test/java/uk/gov/pay/api/resources/directdebit/DirectDebitPaymentsResourceGetIT.java
```diff
@@ -106,6 +107,7 @@ public void searchPayments_success() {
                 .body("results[0].description", Matchers.is(payments.get(0).getDescription()))
                 .body("results[0].reference", Matchers.is(payments.get(0).getReference()))
                 .body("results[0].state.status", Matchers.is(payments.get(0).getState().getStatus()))
+                .body("results[0].state.details", Matchers.is(payments.get(0).getState().getDetails()))
```
oswaldquek: it'd be nice to refactor `Matchers.` to a static import but its irrelevant to this PR i suppose

PR #618, opened 2019-08-01
BAU: replace charge_id with transaction_id 

jankowiakmaria:
* for ledger get one transaction endpoint
* in order to deprecate charge_id


jankowiakmaria:
Thanks!

PR #620, opened 2019-08-01
PP-5495 Remove Events Link From PublicAPI

RoryMalcolm:
- Currently an unused events link is returned inside PublicAPI
responses, as this is currently unused it should be removed from the
builder classes so this confusing information is not returned to users.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/links/directdebit/MandateLinks.java
+++ b/src/main/java/uk/gov/pay/api/model/links/directdebit/MandateLinks.java
@@ -32,7 +32,9 @@ public class MandateLinks {
 
     @JsonProperty(PAYMENTS)
     private final Link payments;
-    
+
+    //Hidden because it is currently unused
+    @ApiModelProperty(hidden = true)
     @JsonProperty(EVENTS)
     private final Link events;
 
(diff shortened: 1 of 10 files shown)


stephencdaly:
Can you make sure the "events" link is removed from the swagger file

RoryMalcolm:
> Can you make sure the "events" link is removed from the swagger file

Done!

swagger/swagger.json
```diff
@@ -1614,6 +1611,10 @@
     "MandateLinks" : {
       "type" : "object",
       "properties" : {
+        "events" : {
+          "description" : "events",
+          "$ref" : "#/definitions/Link"
+        },
```
stephencdaly: It's still here for MandateLinks

src/main/java/uk/gov/pay/api/model/links/directdebit/MandateLinks.java
```diff
@@ -32,7 +32,9 @@
 
     @JsonProperty(PAYMENTS)
     private final Link payments;
-    
+
+    //Hidden because it is currently unused
+    @ApiModelProperty(hidden = true)
```
stephencdaly: I think it's using the @ApiModelProperty on the getter instead, so move the hidden property to there i recon

PR #625, opened 2019-08-06
PP-5418 Behavioural flag

jankowiakmaria:


src/main/java/uk/gov/pay/api/resources/LedgerOrConnectorStrategyTemplate.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(LedgerOrConnectorStrategyTemplate.class);
+    private String strategy;
+    private List<String> VALID_STRATEGIES = ImmutableList.of("ledger-only", "future-behaviour");
+    
+    public LedgerOrConnectorStrategyTemplate(String strategy) {
+        this.strategy = strategy;
+    }
+
+    private void validate() {
+        if(strategy != null && !StringUtils.isEmpty(strategy) && !VALID_STRATEGIES.contains(strategy)) {
```
SandorArpa: !StringUtils.isBlank(strategy) could be used instead of checking for null. Also, isEmpty returns false if there is whitespace only in the String.

src/test/java/uk/gov/pay/api/resources/GetOnePaymentStrategyTest.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
+public class GetOnePaymentStrategyTest {
+    
+    @Mock
+    private GetPaymentService mockGetPaymentService;
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
```
SandorArpa: It's never used

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
```diff
@@ -105,15 +107,16 @@ public PaymentsResource(Client client,
     public Response getPayment(@ApiParam(value = "accountId", hidden = true) @Auth Account account,
                                @PathParam("paymentId")
                                @ApiParam(name = "paymentId", value = "Payment identifier", example = "hu20sqlact5260q2nanm0q8u93")
-                                       String paymentId) {
+                                       String paymentId,
+                               @HeaderParam("X-Ledger") String strategyName) {
```
mrlumbu: maybe deserialise this to an enum, then validation can be done there, and the strategy can just `execute`?

src/test/java/uk/gov/pay/api/resources/GetOnePaymentStrategyTest.java
jankowiakmaria: good spot - was supposed to remove it, but forgot.

src/main/java/uk/gov/pay/api/resources/LedgerOrConnectorStrategyTemplate.java
jankowiakmaria: Sounds good to me - will update.

src/main/java/uk/gov/pay/api/resources/PaymentsResource.java
jankowiakmaria: But then it will throw and I wanted it to just log and fallback to the default strategy...
But maybe it's not a good idea?

mrlumbu: having discussed, this is temporary, so no need to try and do something over fancy. LGTM!

PR #632, opened 2019-08-07
PP-5480 Add dummy card type field

stephencdaly:
Adds a `card_type` field in the `card_details` object in card payment responses. This will currently always return null, but in future will be implemented so that it returns `debit` or `credit` if the card type can be determined.


src/main/java/uk/gov/pay/api/model/CardDetails.java
```diff
@@ -73,4 +76,9 @@ public String getExpiryDate() {
     public String getCardBrand() {
         return cardBrand;
     }
+
+    @ApiModelProperty(value = "The card type, `debit` or `credit` or `null` if not able to determine", allowableValues = "debit,credit,null", example = "debit")
```
oswaldquek: how about `...or null if unknown`?

stephencdaly: "if not able to determine" was what Mark and Alex decided on

alexbishop1: The use of “not able to determine” comes from conversations with @m-green and will match the phrasing in the documentation.

m-green: Just a heads-up that I'm likely to tweak this wording slightly, sorry!

m-green: But I wouldn't let that get in the way of this PR.

stephencdaly: 👍

PR #637, opened 2019-08-08
PP-5421 Refactor PaymentRefunds resource method

kbottla:
## WHAT YOU DID

_Prep to use Ledger for PaymentRefunds_
  - Refactored PaymentRefunds resource method to use `GetPaymentRefundsService` to get refunds from connector

### DIFF
--- a/src/main/java/uk/gov/pay/api/service/ConnectorService.java
+++ b/src/main/java/uk/gov/pay/api/service/ConnectorService.java
@@ -1,9 +1,13 @@
 package uk.gov.pay.api.service;
 
+import uk.gov.pay.api.app.config.PublicApiConfig;
 import uk.gov.pay.api.auth.Account;
 import uk.gov.pay.api.exception.GetChargeException;
+import uk.gov.pay.api.exception.GetRefundsException;
 import uk.gov.pay.api.model.Charge;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/api/service/ConnectorService.java
```diff
@@ -34,4 +42,18 @@ public Charge getCharge(Account account, String paymentId) {
...
+
+    public RefundsResponse getPaymentRefunds(Account account, String paymentId) {
+        Response connectorResponse = client
+                .target(connectorUriGenerator.refundsForPaymentURI(account.getAccountId(), paymentId))
+                .request()
+                .get();
+
+        if (connectorResponse.getStatus() == SC_OK) {
+            RefundsFromConnector refundsFromConnector = connectorResponse.readEntity(RefundsFromConnector.class);
+            return RefundsResponse.valueOf(refundsFromConnector, baseUrl);
```
jankowiakmaria: I don't think `ConnectorService` should know about `publicApiConfig`.
It should return `RefundsFromConnector` and map it in `GetPaymentRefundsService` which is higher lever class.

kbottla: Agree! It will move with further refactoring

PR #638, opened 2019-08-08
PP-5395 get payment events from ledger strategy

jankowiakmaria:
* Moved getChargeEvents to ConnectorService
* Introduced ledger endpoint
* Added strategy for get payment events
* Added strategy test
* Added contract test for getting payment events from ledger
* Cleanup

**Note**
ledger pact test is disabled (by disabled I mean deleted) - will be enabled (reintroduced from the first commit of this PR) as soon as I prepare relevant code on ledger side


src/main/java/uk/gov/pay/api/service/ConnectorService.java
```diff
@@ -43,6 +46,22 @@ public Charge getCharge(Account account, String paymentId) {
...
 
+    public PaymentEvents getChargeEvents(Account account, String paymentId) {
+        Response connectorResponse = client
+                .target(connectorUriGenerator.chargeEventsURI(account, paymentId))
+                .request()
+                .accept(MediaType.APPLICATION_JSON)
+                .get();
+
+        if (connectorResponse.getStatus() == SC_OK) {
+            PaymentEvents response = connectorResponse.readEntity(PaymentEvents.class);
```
kbottla: Not a big thing but we can return response directly here

src/test/java/uk/gov/pay/api/service/GetPaymentEventsServiceTest.java
```diff
@@ -28,30 +29,36 @@
     @Rule
     public PactProviderRule connectorRule = new PactProviderRule("connector", this);
 
+    @Rule
+    public PactProviderRule ledgerRule = new PactProviderRule("ledger", this);
```
kbottla: I am guessing there will be pact test that checks ledger events

jankowiakmaria: precisely (if you look at the first commit only you can even see it - I'll reintroduce it in the next PR).

src/main/java/uk/gov/pay/api/service/ConnectorService.java
jankowiakmaria: yeah, even IDE told me that, but in the end I forgot to do it. Will update separately.

PR #639, opened 2019-08-08
PP-5421 Refactor GetPaymentRefundsService

kbottla:
## WHAT 

Prep to use Ledger to get Refunds for a payment

- Refactored `GetPaymentRefundsService` to do more processing (building response with links) instead of delegating to ConnectorService
- `ConnectorService` now returns a generic `Refund` object for payment refunds
- PaymentRefundsResource returns concrete object (where possible) instead of building response with `Response`
- `RefundsResponse` and `RefundResponse` now represent the actual response for refunds instead of using `HalResourceResponse` to build response
- Removed `HalResourceResponse` as it is not longer needed


src/main/java/uk/gov/pay/api/service/GetPaymentRefundsService.java
```diff
@@ -1,20 +1,64 @@
...
 import uk.gov.pay.api.model.RefundsResponse;
 
 import javax.inject.Inject;
+import javax.ws.rs.core.UriBuilder;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static uk.gov.pay.api.resources.PaymentRefundsResource.PAYMENT_BY_ID_PATH;
+import static uk.gov.pay.api.resources.PaymentRefundsResource.PAYMENT_REFUNDS_PATH;
+import static uk.gov.pay.api.resources.PaymentRefundsResource.PAYMENT_REFUND_BY_ID_PATH;
```
jankowiakmaria: I don't like the fact that service that is used by the resource imports static things from this resource (sounds like circular dependency to me). Should be moved somewhere else (follow up PR I guess)

kbottla: Good spot! will replace in next PR

PR #642, opened 2019-08-09
PP-4769- Adds null check for PaymentOutcomeValidator to handle missing status

nimalank7:
This PR adds a null check in PaymentOutcomeValidator to ensure that the exception is handled when no status is provided in the PaymentOutcome section of the JSON. I have added 2 tests in the PaymentOutcomeValidationResourceTest to check this.

`"payment_outcome": { 
  }`


nimalank7:
Rebased

PR #647, opened 2019-08-12
PP-5421 Pact test for ledger payment refunds

kbottla:


src/test/java/uk/gov/pay/api/service/GetPaymentRefundsServiceTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+        assertThat(refunds.get(0).getRefundId(), is("refund-transaction-id1"));
+        assertThat(refunds.get(0).getStatus(), is("success"));
+        assertThat(refunds.get(0).getAmount(), is(1000L));
+        assertThat(refunds.get(0).getCreatedDate(), is("2018-09-22T10:14:16.067Z"));
+        assertThat(refunds.get(0).getLinks().getPayment().getHref(), is("http://publicapi.test.localhost/v1/payments/ch_123abc456xyz"));
+        assertThat(refunds.get(0).getLinks().getSelf().getHref(), is("http://publicapi.test.localhost/v1/payments/ch_123abc456xyz/refunds/refund-transaction-id1"));
+
+        assertThat(refunds.get(1).getRefundId(), is("refund-transaction-id2"));
+        assertThat(refunds.get(1).getStatus(), is("success"));
```
jankowiakmaria: Would be good to have unsuccessful one as well.

src/test/java/uk/gov/pay/api/service/GetPaymentRefundsServiceTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+        Client client = RestClientFactory.buildClient(new RestClientConfig(false));
+        LedgerService ledgerService = new LedgerService(client, ledgerUriGenerator);
+
+        getPaymentRefundsService = new GetPaymentRefundsService(mockConnectorService, ledgerService, publicApiUriGenerator);
+    }
+
+    @Test
+    @PactVerification("ledger")
+    @Pacts(pacts = {"publicapi-ledger-get-payment-refunds"})
```
jankowiakmaria: I'm not convinced by the name (of the file).
Shouldn't it be more like `publicapi-ledger-get-transaction-refunds` or
`publicapi-ledger-get-transaction-transaction` (this one doesn't sound great either though).

kbottla: This naming makes it clear of what we are testing (and the action) from publicapi point of view. Ledger state is named `refund transactions for a transaction exist` which makes it relevant within ledger

We follow similar naming convention for all ledger and connector pacts

src/test/java/uk/gov/pay/api/service/GetPaymentRefundsServiceTest.java
kbottla: 👍 Will update ledger state and then make changes here.

kbottla: Updated https://github.com/alphagov/pay-publicapi/blob/149e2c833dc3e0af77128f355fb4e27ddf20cd20/src/test/resources/pacts/publicapi-ledger-get-payment-refunds.json#L50

PR #649, opened 2019-08-12
PP-5519 Helpful Error When MandateID Invalid

RoryMalcolm:
- Returns a helpful error when the MandateID is invalid


src/main/java/uk/gov/pay/api/model/PaymentError.java
```diff
@@ -51,6 +51,8 @@
         CREATE_PAYMENT_REFUND_AMOUNT_AVAILABLE_MISMATCH("P0604", "Refund amount available mismatch."),
         CREATE_PAYMENT_MANDATE_STATE_INVALID("P1103",
                 "Can't create payment. The mandate's state must be pending or active."),
+        CREATE_PAYMENT_MANDATE_ID_INVALID("P1100",
+                "Can't create payment. The mandate with the corresponding ID does not exist."),
```
danworth: I think it would helpful to reference the id in the message, something as per the story `A mandate with id xxxx was not found`

PR #650, opened 2019-08-12
PP-5498: add strategy for get refund

jankowiakmaria:
* created GetPaymentRefundStrategy + test
* created GetPaymentRefundService
* moved retrieving refund from Connector to ConnectorService
* added pact test for connector

Todo (outside of this PR):
* rename `TransactionResponse` to `TransactionFromLedger` (or similar)
  * Refunds should use it as well, that would allow a little bit of clean-up later on
* use transaction_type for getting payment as well (ideally merge two methods from LedgerService)
* pact test for ledger

### DIFF
--- a/src/main/java/uk/gov/pay/api/service/LedgerService.java
+++ b/src/main/java/uk/gov/pay/api/service/LedgerService.java
@@ -3,6 +3,8 @@ package uk.gov.pay.api.service;
 import uk.gov.pay.api.auth.Account;
 import uk.gov.pay.api.exception.GetChargeException;
 import uk.gov.pay.api.exception.GetEventsException;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/pay/api/service/LedgerService.java
```diff
@@ -3,6 +3,8 @@
 import uk.gov.pay.api.auth.Account;
 import uk.gov.pay.api.exception.GetChargeException;
 import uk.gov.pay.api.exception.GetEventsException;
+import uk.gov.pay.api.exception.GetRefundException;
```
kbottla: Unused import

PR #651, opened 2019-08-12
PP-5421 Pact test for connector payment refunds

kbottla:
## WHAT YOU DID
- Adds pact tests for payment refunds from connector

### DIFF
--- a/src/test/resources/pacts/publicapi-connector-get-payment-refunds.json
+++ b/src/test/resources/pacts/publicapi-connector-get-payment-refunds.json
@@ -1,0 +1,210 @@
+{
+  "consumer": {
+    "name": "publicapi"
+  },
+  "provider": {
+    "name": "connector"
+  },
+  "interactions": [
+    {
+      "description": "Get refunds for a payment",
+      "providerStates": [
+        {
+          "name": "Refunds exist for a charge",
+          "params": {
+            "account_id": "123456",
+            "charge_id": "charge8133029783750222"
+          }
+        }
+      ],
+      "request": {
+        "method": "GET",
+        "path": "/v1/api/accounts/123456/charges/charge8133029783750222/refunds"
+      },
+      "response": {
+        "status": 200,
+        "headers": {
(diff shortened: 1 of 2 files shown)


src/test/resources/pacts/publicapi-connector-get-payment-refunds.json
```diff
@@ -0,0 +1,210 @@
...
+              "matchers": [
+                {
+                  "regex": "http.*://.*/v1/api/accounts/123456/charges\/[a-z0-9]*\/refunds\/[a-z0-9]*"
+                }
+              ]
+            },
+            "$._embedded.refunds[0]._links.payment.href": {
+              "matchers": [
+                {
+                  "regex": "http.*://.*/v1/api/accounts/123456/charges/charge8133029783750222"
```
jankowiakmaria: Do we want to check whether it's the exact value or rather match it with `\/[a-z0-9]*\`? (same for account id)

src/test/resources/pacts/publicapi-connector-get-payment-refunds.json
```diff
@@ -0,0 +1,210 @@
...
+              "matchers": [
+                {
+                  "match": "type"
+                }
+              ]
+            },
+            "$._embedded.refunds[0].created_date": {
+              "matchers": [
+                {
+                  "match": "value"
```
jankowiakmaria: Should it be value, type or datetime pattern in the case of `created_date`?

kbottla: I think value is fine. Not exactly sure why we have patterns in some places, but one reason I can think of is createdDate set to ZonedDateTime.now() in states which makes it hard to check for value

src/test/resources/pacts/publicapi-connector-get-payment-refunds.json
kbottla: Values should be fine here as they won't change

PR #654, opened 2019-08-13
PP-5499 Refactor SearchRefundsService

kbottla:
## WHAT YOU DID

Prep to introduce ledger for refunds search

- Refactors SearchRefundsService to use ConnectorService to search for refunds
- Returns an object instead of generic `Response` fron SearchRefundsResource

### DIFF
--- a/src/main/java/uk/gov/pay/api/service/ConnectorService.java
+++ b/src/main/java/uk/gov/pay/api/service/ConnectorService.java
@@ -5,16 +5,19 @@ import uk.gov.pay.api.exception.GetChargeException;
 import uk.gov.pay.api.exception.GetEventsException;
 import uk.gov.pay.api.exception.GetRefundException;
 import uk.gov.pay.api.exception.GetRefundsException;
+import uk.gov.pay.api.exception.SearchRefundsException;
 import uk.gov.pay.api.model.Charge;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/api/service/ConnectorService.java
```diff
@@ -82,4 +85,17 @@ public RefundFromConnector getPaymentRefund(String accountId, String paymentId,
...
+
+    public SearchRefundsResponseFromConnector searchRefunds(Account account, Map<String, String> queryParams) {
+        Response connectorResponse = client
+                .target(connectorUriGenerator.refundsURIWithParams(account, queryParams))
+                .request()
+                .accept(MediaType.APPLICATION_JSON)
+                .get();
+
+        if (connectorResponse.getStatus() == SC_OK) {
+            return connectorResponse.readEntity(SearchRefundsResponseFromConnector.class);
```
mrlumbu: think you've removed the `try/catch` for `ProcessingException` here?

kbottla: Good catch! Updated here https://github.com/alphagov/pay-publicapi/pull/654/commits/ad9beb687556b75d7f754f129e5fec70cd356b6a#diff-07acbd73c1ac179bb38becfe7d2de74bR98

PR #670, opened 2019-08-28
bau: Make swagger doc on metadata clearer

oswaldquek:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.api.model;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 import org.hibernate.validator.constraints.Length;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
```diff
@@ -136,6 +137,8 @@ public String getReturnUrl() {
     }
 
     @JsonProperty("metadata")
+    @ApiModelProperty(value = "Additional metadata - up to 10 key-value pairs - on the payment. Each key must be between 1 and 30 characters long. Values must be no greater than 50 characters long and be of types String, Boolean or Number.",
```
alexbishop1: The 50 character limit only applies if it’s a string value. Also, https://json.org/ calls them “name/value” pairs (rather than “key-value” pairs, so perhaps we should use that terminology and apparently “string” and “number” and the like don’t start with capital letters in JavaScriptland. Might be useful to mention it’s an object too. So maybe tweak it to be something like this: “Additional metadata — object containing up to 10 name/value pairs. Each name must be between 1 and 30 characters long. Each value must be either a string (no greater than 50 characters long), boolean or number.”

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
```diff
@@ -136,6 +137,8 @@ public String getReturnUrl() {
...
+            dataType = "java.util.Map", example = "{'ledger_code':'123', 'reconciled', true}")
```
alexbishop1: JSON ~keys~ names have to be in double quotes and I think that comma after “reconciled” should be a boolean.

PR #674, opened 2019-08-29
PP-5607 RateLimiterFilter - Improve logging

kbottla:
## WHAT YOU DID

- Updated RateLimiterFilter type to ContainerRequestFilter (from Filter) so security context is available for the filter (that holds `Account` principal object)
- Assigned priority `@Priority(Priorities.USER)` so that the filter comes after `OAuthCredentialAuthFilter`(dropwizard) which authorises token and creates security context
- RedisRateLimiter & LocalRateLimiter now logs account ID and HTTP method if requests are rate limited
- Updated relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
+++ b/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
@@ -67,7 +67,7 @@ public class RateLimiterFilter implements ContainerRequestFilter {
 
         String accountId = getAccountId(requestContext);
         try {
-            rateLimiter.checkRateOf(method + "-" + authorization, method);
+            rateLimiter.checkRateOf(accountId, method + "-" + authorization, method);
         } catch (RateLimitException e) {
             LOGGER.info("Rate limit reached for current service and account [{}]. Sending response '429 Too Many Requests'", accountId);
             setTooManyRequestsError();
--- a/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
+++ b/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
@@ -69,7 +69,7 @@ public class RateLimiterFilter implements ContainerRequestFilter {
         try {
             rateLimiter.checkRateOf(accountId, method + "-" + authorization, method);
         } catch (RateLimitException e) {
-            LOGGER.info("Rate limit reached for current service and account [{}]. Sending response '429 Too Many Requests'", accountId);
+            LOGGER.info("Rate limit reached for current service [account - {}, method - {}]. Sending response '429 Too Many Requests'", accountId, method);
             setTooManyRequestsError();
         }
     }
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
```diff
@@ -43,31 +50,34 @@ public RateLimiterFilter(RateLimiter rateLimiter, ObjectMapper objectMapper) {
...
+        String accountId = getAccountId(requestContext);
         try {
-            rateLimiter.checkRateOf(method + "-" + authorization, method);
-            chain.doFilter(request, response);
+            rateLimiter.checkRateOf(accountId, method + "-" + authorization, method);
         } catch (RateLimitException e) {
-            LOGGER.info("Rate limit reached for current service. Sending response '429 Too Many Requests'");
-            setTooManyRequestsError((HttpServletResponse) response);
-        } 
+            LOGGER.info("Rate limit reached for current service and account [{}]. Sending response '429 Too Many Requests'", accountId);
```
danworth: Looks like we check the rate limiting based on a composite key comprised of the method and the api key. Makes sense not to log the api key but should this log entry mention the method along with the account id? I see we log the method as part of throwing the `RateLimiteException` so perhaps not necessary to add it here too, so happy to approve as-is 👍

src/test/java/uk/gov/pay/api/filter/RateLimiterFilterTest.java
```diff
@@ -2,80 +2,73 @@
...
     @Test
-    public void shouldProcessFilterChain_whenRequestsAreWithinTheRate() throws Exception {
-        rateLimiterFilter.doFilter(mockPostRequest, mockResponse, mockFilterChain);
-        rateLimiterFilter.doFilter(mockGetRequest, mockResponse, mockFilterChain);
-
-        verify(mockFilterChain, times(1)).doFilter(mockPostRequest, mockResponse);
-        verify(mockFilterChain, times(1)).doFilter(mockGetRequest, mockResponse);
+    public void shouldCheckRateLimitsWhenFilterIsInvoked() throws Exception {
+        rateLimiterFilter.filter(mockContainerRequestContext);
+        verify(rateLimiter).checkRateOf(any(), any(), any());
```
danworth: This could be a bit tighter as I think it should be called with `account-id` and the `mockSecurityContext` could be made to return known values for the api key and method that could be checked too? That might help assure that the correct values are being taken from the security context?

kbottla: 👍 Updated

src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
kbottla: Added `method` for consistency

PR #677, opened 2019-08-30
BAU fix CreatePaymentResult swagger model

SandorArpa:
## WHAT YOU DID
- swagger model is out of date, missing (optional) fields, like email and card_details
- see https://govuk.zendesk.com/agent/tickets/3783443

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CreatePaymentResult.java
+++ b/src/main/java/uk/gov/pay/api/model/CreatePaymentResult.java
@@ -72,4 +72,12 @@ public class CreatePaymentResult {
     @JsonProperty
     @ApiModelProperty(name = "metadata", dataType = "Map[String,String]")
     private ExternalMetadata metadata;
+
+    @JsonProperty
+    @ApiModelProperty(name = "email", example = "citizen@example.org", required = false)
+    private String email;
+
+    @JsonProperty(value = "card_details")
+    @ApiModelProperty(name = "card_details", required = false)
+    private CardDetails cardDetails;
 }
--- a/swagger/swagger.json
+++ b/swagger/swagger.json
@@ -1299,11 +1299,18 @@
             "type" : "string"
           }
         },
+        "email" : {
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/api/model/CreatePaymentResult.java
```diff
@@ -72,4 +72,11 @@
     @JsonProperty
     @ApiModelProperty(name = "metadata", dataType = "Map[String,String]")
     private ExternalMetadata metadata;
+
+    @JsonProperty
+    @ApiModelProperty(name = "email", example = "citizen@example.org")
```
kbottla: Could be useful to indicate that the fields are not mandatory
```suggestion
    @ApiModelProperty(name = "email", example = "citizen@example.org", required=false)
```

src/main/java/uk/gov/pay/api/model/CreatePaymentResult.java
```diff
@@ -72,4 +72,11 @@
...
+    private String email;
+
+    @JsonProperty("card_details")
+    private CardDetails cardDetails;
```
kbottla: Same here to indicate it is not mandatory.

src/main/java/uk/gov/pay/api/model/CreatePaymentResult.java
```diff
@@ -72,4 +72,11 @@
     @JsonProperty
     @ApiModelProperty(name = "metadata", dataType = "Map[String,String]")
     private ExternalMetadata metadata;
+
+    @JsonProperty
+    @ApiModelProperty(name = "email", example = "citizen@example.org", required = false)
+    private String email;
+
+    @JsonProperty(value = "card_details", required = false)
```
kbottla:
```suggestion
    @JsonProperty(value = "card_details")
    @ApiModelProperty(name = "card_details", required = false)
```

PR #681, opened 2019-09-03
PP-5612 Add gateway account id to logs

danworth:
PP-5612 Add gateway account id to logs
    
    To aid debugging this adds the gateway account id to the MDC using a
    `ContainerRequestFilter` so it can be logged in subsequent lines. The value is
    cleared by  a new `ClearMdcValuesFilter` class which is a
    `ServletFilter` at the head of the filter chain, this allows it to clean the
    mdc values within a `finally` block after calling `chain.doFilter()`.
    
    An alternative to using the `ClearMdcValuesFilter` to clean up at the
    end of the request would be follow the pattern of the `LoggingFilter`
    which we use to add the request-id. This will replace the request-id if
    it already exists, else set it a new. Although this seems simpler it
    has the risk of any log lines printed prior to `LoggingFilter`
    containing a stale request-id; instead it seems safer to clean the mdc
    values once the request has completed and the thread released to the
    pool.
    
    Ultimately it seems dealing with MDC and threadpools is a little tricky
    but we think this is a reasonable solution.

## How to test
Checked this by looking at the log output from IT and checking when the gateway account id appears and is removed.

Worth noting that this logs the gateway account id in the event of a 422 response (which triggered the need to make this change).
```
[2019-09-03 10:55:29.320] [thread=dw-62] [level=INFO ] [logger=u.g.p.c.u.l.LoggingFilter] [requestID=(none)] [gateway_account_id=GATEWAY_ACCOUNT_ID] - POST to /v1/payments/ ended - total time 233ms
127.0.0.1 - - [03/Sep/2019:09:55:29 +0000] "POST /v1/payments/ HTTP/1.1" 422 117 "-" "Apache-HttpClient/4.5.9 (Java/11)" 243
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
         <jackson.version>2.9.9</jackson.version>
         <logback.version>1.2.3</logback.version>
         <docker-client.version>8.16.0</docker-client.version>
-        <pay-java-commons.version>1.0.20190816103843</pay-java-commons.version>
+        <pay-java-commons.version>1.0.20190828103217</pay-java-commons.version>
         <pact.version>3.6.12</pact.version>
         <PACT_BROKER_URL/>
         <PACT_BROKER_USERNAME/>
@@ -51,6 +51,11 @@
         <dependency>
             <groupId>uk.gov.pay</groupId>
             <artifactId>validation</artifactId>
+            <version>${pay-java-commons.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>uk.gov.pay</groupId>
+            <artifactId>logging</artifactId>
             <version>${pay-java-commons.version}</version>
         </dependency>
         <dependency>
--- a/src/main/java/uk/gov/pay/api/app/PublicApi.java
+++ b/src/main/java/uk/gov/pay/api/app/PublicApi.java
@@ -41,6 +41,8 @@ import uk.gov.pay.api.exception.mapper.SearchChargesExceptionMapper;
 import uk.gov.pay.api.exception.mapper.SearchRefundsExceptionMapper;
 import uk.gov.pay.api.exception.mapper.ViolationExceptionMapper;
 import uk.gov.pay.api.filter.AuthorizationValidationFilter;
+import uk.gov.pay.api.filter.ClearMdcValuesFilter;
+import uk.gov.pay.api.filter.GatewayAccountIdFilter;
 import uk.gov.pay.api.filter.RateLimiterFilter;
 import uk.gov.pay.api.healthcheck.Ping;
 import uk.gov.pay.api.ledger.resource.TransactionsResource;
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/api/filter/AuthorizationValidationFilter.java
```diff
@@ -38,8 +40,11 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         final String authorization = ((HttpServletRequest) request).getHeader("Authorization");
 
         if (isValidAuthorizationHeader(authorization)) {
-            chain.doFilter(request, response);
-
+            try {
+                chain.doFilter(request, response);
+            } finally {
+                MDC.remove(GATEWAY_ACCOUNT_ID);
```
rhowe-gds: This seems rather incongruent to me - should we define a dedicated filter for managing request-lifecycle MDC keys?

danworth: Yeah I think thats a reasonable suggestion. We could create an MDC Filter and give it priority so it is at the edge of our app and clear the MDC values there. 
```java
try {
        chain.doFilter(request, response);
} finally {
         MDC.clear();
}
```

I'll make the change 👍

src/main/java/uk/gov/pay/api/auth/AccountAuthenticator.java
```diff
@@ -55,6 +57,7 @@ public AccountAuthenticator(Client client, PublicApiConfig configuration) {
     private Optional<Account> readAccountFromResponse(Response response) {
         JsonNode responseEntity = response.readEntity(JsonNode.class);
         String accountId = responseEntity.get("account_id").asText();
+        MDC.put(GATEWAY_ACCOUNT_ID, accountId);
```
kbottla: This might not get invoked all the time. `AccountAuthenticator` result is cached by Dropwizard, so AuthFilter might not invoke `AccountAuthenticator. readAccountFromResponse` if object still exists in cache (but not exactly sure of the behaviour)
https://github.com/alphagov/pay-publicapi/blob/master/src/main/java/uk/gov/pay/api/app/PublicApi.java#L120

We can probably create MdcFilter as a ContainerRequestFilter (which we can set to be invoked after Authentication) and it can both put and clear `gateway_account_id`.

danworth: Its a great point about the caching and therefor we cannot set the value here. As discussed I don't think we can use a `ContainerRequestFilter` to tidy up and call `MDC.clean()` since it has no means to be called back when the request is complete. Looking at `LoggingFilter` it will remove the request-id mdc value if present before setting it. This seems a pragmatic approach but could lead to some confusion if anything is called prior to that filter as it would have a stale value. Perhaps a separate container filter to add it with `USER` priority and another servlet filter at the head of the chain to clean all mdc values...

kbottla: Is it still required here ?

danworth: I'd just realised that I'd left this in, I've removed and pushed.

PR #686, opened 2019-09-07
PP-5443 Integration Test infrastructure for Telephone payments

nimalank7:
## WHAT YOU DID
This PR adds the appropriate methods in ConnectorMockClient for mocking out Connector's response for telephone payments (e.g. 200 and 201).

- It renames TelephoneChargeBuilder to Builder in CreateTelephonePaymentRequest. 
- It also modifies PaymentOutcomeValidatorTest to directly test the constraint violation message for not-null.
- Removes ChargeServiceTest 
- Removes State and refactors CreateTelephonePaymentRequest and CreateTelephonePaymentResponse to use PaymentState instead
- Adds appropriate Jackson annotations to properly handle serializing/deserializing optional fields.


src/main/java/uk/gov/pay/api/model/telephone/CreateTelephonePaymentRequest.java
```diff
@@ -23,24 +27,32 @@
...
     @Size(max = REFERENCE_MAX_LENGTH, message = "Must be less than or equal to {max} characters length")
+    @NotNull(message = "Field [reference] cannot be null")
     private String reference;
     
     @Size(max = DESCRIPTION_MAX_LENGTH, message = "Must be less than or equal to {max} characters length")
+    @NotNull(message = "Field [description] cannot be null")
     private String description;
-    
+
+    @JsonProperty("created_date")
```
mrlumbu: the `SnakeCaseStrategy` should mean you don't need to add these `JsonProperty` annotations, unless I'm missing something?

nimalank7: For the Optionals to work with Jackson we have to add '@JsonProperty...'

PR #697, opened 2019-09-17
PP-5632: add configuration for rate-limiting

jankowiakmaria:
## Why
In order to support extended limits for chosen accounts we need to extend configuration.

## What
* add additional parameter values for rate-limitting
* introduce converter to transform string into list of strings
* cover changes with tests
* update README

### DIFF
--- a/src/main/java/uk/gov/pay/api/app/config/StringToListConverter.java
+++ b/src/main/java/uk/gov/pay/api/app/config/StringToListConverter.java
@@ -1,0 +1,21 @@
+package uk.gov.pay.api.app.config;
+
+import com.fasterxml.jackson.databind.util.StdConverter;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/api/app/config/StringToListConverter.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class StringToListConverter extends StdConverter<String, List<String>> {
+
+    @Override
+    public List<String> convert(String value) {
+        if (StringUtils.isBlank(value)) {
+            return Collections.emptyList();
+        }
+
+        return Arrays.stream(value.split(",")).map(String::trim).collect(Collectors.toList());
```
kbottla: May be also filter by nonEmpty string to avoid list with empty strings for input like `a,,b,`

jankowiakmaria: Sure, sounds good. And it couldn't be done as separate PR, because?

jankowiakmaria:
Will address those two points separately.
Thanks!

PR #701, opened 2019-09-18
PP-5632: Implementation for rate-limiting

jankowiakmaria:
Context:
At the moment rate limiting works on per method basis [GET/POST].
It is required to introduce separate buckets (each bucket will have separate limit counter):
* [POST] create payment
* [POST] capture payment
* remaining [POST] requests
* all [GET] requests

In order to do that the RateLimiterKey has been introduces. It determines the bucket
based on the method and path (taken from ContainerRequestContext).
On top of splitting endpoints into separate buckets the logic to have two types of limits
has been introduced. Certain accounts (defined in configuration #697) will have different limits
than by default. In the implementation RateLimitManager is used to determine the limit based on
account id and RateLimiterKey (although at the moment only method from the object is used for the limit).

The changes affect both limiters (redis and local one) although local one is affected only by bucketing.

Changes:
* introduce PathHelper that returns path type depending on path and method
* introduce RateLimiterKey that based on ContainerRequestContext constructs itself (key, keyType, method)
  it extends the previous way of generating the request key
* introduce RateLimitManager which returns the allowed number of requests for
given rateLimiterKey and account

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
+++ b/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
@@ -62,12 +62,12 @@ public class RateLimiterFilter implements ContainerRequestFilter {
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
-        final String authorization = requestContext.getHeaderString("Authorization");
         final String method = requestContext.getMethod();
 
         String accountId = getAccountId(requestContext);
+        RateLimiterKey key = RateLimiterKey.from(requestContext);
         try {
-            rateLimiter.checkRateOf(accountId, method + "-" + authorization, method);
+            rateLimiter.checkRateOf(accountId, key, method);
         } catch (RateLimitException e) {
             LOGGER.info("Rate limit reached for current service [account - {}, method - {}]. Sending response '429 Too Many Requests'", accountId, method);
             setTooManyRequestsError();
--- a/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
+++ b/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
@@ -65,9 +65,9 @@ public class RateLimiterFilter implements ContainerRequestFilter {
         final String method = requestContext.getMethod();
 
         String accountId = getAccountId(requestContext);
-        RateLimiterKey key = RateLimiterKey.from(requestContext);
+        RateLimiterKey key = RateLimiterKey.from(requestContext, accountId);
         try {
-            rateLimiter.checkRateOf(accountId, key, method);
+            rateLimiter.checkRateOf(accountId, key);
         } catch (RateLimitException e) {
             LOGGER.info("Rate limit reached for current service [account - {}, method - {}]. Sending response '429 Too Many Requests'", accountId, method);
             setTooManyRequestsError();
--- a/src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
+++ b/src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
@@ -1,0 +1,46 @@
+package uk.gov.pay.api.filter;
+
+import uk.gov.pay.api.utils.PathHelper;
+
+import javax.ws.rs.container.ContainerRequestContext;
+
+public class RateLimiterKey {
+    
+    private String method;
+    private String key;
+    private String keyType;
+
+    public RateLimiterKey(String key, String keyType, String method) {
+        this.key = key;
+        this.keyType = keyType;
+        this.method = method;
+    }
+
+    public static RateLimiterKey from(ContainerRequestContext requestContext) {
+        final String method = requestContext.getMethod();
+
+        StringBuilder builder = new StringBuilder(method);
+
+        final String pathType = PathHelper.getPathType(requestContext.getUriInfo().getPath(), method);
+        if (!pathType.isBlank()) {
+            builder.append("-" + pathType);
+        }
+
+        final String keyType = builder.toString();
+        builder.append("-" + requestContext.getHeaderString("Authorization"));
+
+        return new RateLimiterKey(builder.toString(), keyType, method);
+    }
+
+    public String getKey() {
+        return key;
+    }
+
+    public String getKeyType() {
+        return keyType;
+    }
+
+    public String getMethod() {
+        return method;
+    }
+}
--- a/src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
+++ b/src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
@@ -1,8 +1,11 @@
 package uk.gov.pay.api.filter;
 
+import org.apache.commons.lang3.StringUtils;
+import uk.gov.pay.api.auth.Account;
 import uk.gov.pay.api.utils.PathHelper;
 
 import javax.ws.rs.container.ContainerRequestContext;
+import java.util.Optional;
 
 public class RateLimiterKey {
     
@@ -16,7 +19,7 @@ public class RateLimiterKey {
         this.method = method;
     }
 
-    public static RateLimiterKey from(ContainerRequestContext requestContext) {
+    public static RateLimiterKey from(ContainerRequestContext requestContext, String accountId) {
         final String method = requestContext.getMethod();
 
         StringBuilder builder = new StringBuilder(method);
@@ -27,7 +30,7 @@ public class RateLimiterKey {
         }
 
         final String keyType = builder.toString();
-        builder.append("-" + requestContext.getHeaderString("Authorization"));
+        builder.append("-" + accountId);
 
         return new RateLimiterKey(builder.toString(), keyType, method);
     }
--- a/src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimitManager.java
+++ b/src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimitManager.java
@@ -1,0 +1,29 @@
+package uk.gov.pay.api.filter.ratelimit;
+
+import uk.gov.pay.api.app.config.RateLimiterConfig;
+import uk.gov.pay.api.filter.RateLimiterKey;
+
+public class RateLimitManager {
+   
+    private RateLimiterConfig configuration;
+
+    public RateLimitManager(RateLimiterConfig config) {
+        configuration = config;
+    }
+
+    public int getAllowedNumberOfRequests(RateLimiterKey rateLimiterKey, String account) {
+        if(configuration.getElevatedAccounts().contains(account)) {
+            if("POST".equals(rateLimiterKey.getMethod())) {
+                return configuration.getNoOfPostReqForElevatedAccounts();
+            }
+
+            return configuration.getNoOfReqForElevatedAccounts();
+        }
+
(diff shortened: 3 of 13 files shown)


jankowiakmaria:
There is one comment from @heathd on the commit itself: https://github.com/alphagov/pay-publicapi/commit/d986e67220728b4ef0fc40c0f2cc47d2f4445b3e#commitcomment-35121994

src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
```diff
@@ -0,0 +1,46 @@
...
+
+        StringBuilder builder = new StringBuilder(method);
+
+        final String pathType = PathHelper.getPathType(requestContext.getUriInfo().getPath(), method);
+        if (!pathType.isBlank()) {
+            builder.append("-" + pathType);
+        }
+
+        final String keyType = builder.toString();
+        builder.append("-" + requestContext.getHeaderString("Authorization"));
```
mrlumbu: I realise you didn't introduce this idea, but I'm not sure about using the raw API key as part of the redis key? Can we not make a cheap hash of it ore something?

mrlumbu: a further thought - it looks like you have access to the account id when creating the rate limit key, so why not use that rather than api key?

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -17,13 +18,13 @@ public RateLimiter(LocalRateLimiter localRateLimiter, RedisRateLimiter redisRate
...
-    public void checkRateOf(String accountId, String key, String method) throws RateLimitException {
+    public void checkRateOf(String accountId, RateLimiterKey key, String method) throws RateLimitException {
         try {
-            redisRateLimiter.checkRateOf(accountId, key, method);
+            redisRateLimiter.checkRateOf(accountId, key);
         } catch (RedisException e) {
             LOGGER.warn("Exception occurred checking rate limits using RedisRateLimiter, falling back to LocalRateLimiter");
 
-            localRateLimiter.checkRateOf(accountId, key, method);
+            localRateLimiter.checkRateOf(accountId, key.getKey(), method);
```
mrlumbu: `localRateLimiter.checkRateOf(accountId, key.getKey(), key.getMethod());`?
Then you don't need to pass in method into this method (if you se what I mean)

src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
jankowiakmaria: 🤷‍♀ 
I didn't want to change the existing logic. Happy to change to account id if that's preferable.

src/test/java/uk/gov/pay/api/filter/ratelimit/RateLimitManagerTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @Test
+    @Parameters({
+            "1,POST,4",
+            "1,GET,3",
+            "2,POST,2",
+            "2,GET,1",
+            ",PUT,1"
+    })
+    public void returnsNumberOfAllowedRequests(String account, String method, int expectedNumberOfAllowedRequests) {
+        when(rateLimiterKey.getMethod()).thenReturn(method);
```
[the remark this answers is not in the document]
jankowiakmaria: I didn't care whether it's a real object or a mock, given the test is about `RateLimitManager`.
Using a mock clearly indicates that it's not about `RateLimiterKey`.
I can update. I don't **really** have a strong preference.

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
jankowiakmaria: Oh yeah, fair enough. I've added `method` later and forgot about it in the clean-up phase.

src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
mrlumbu: @kbottla any thoughts on this - was there a reason to use api key rather than account id?

src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
```diff
@@ -62,12 +62,12 @@ private void setTooManyRequestsError() throws IOException {
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
-        final String authorization = requestContext.getHeaderString("Authorization");
         final String method = requestContext.getMethod();
 
         String accountId = getAccountId(requestContext);
+        RateLimiterKey key = RateLimiterKey.from(requestContext, accountId);
```
jankowiakmaria: I know I could calculate it in the RateLimiterKey, but that would introduce the whole lot of changes (it can be done separately though)

src/test/java/uk/gov/pay/api/filter/ratelimit/RateLimitManagerTest.java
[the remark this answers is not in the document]
jankowiakmaria: Updated

src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
jankowiakmaria: I've changed it to account id (in the second commit)

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
jankowiakmaria: Updated

src/test/java/uk/gov/pay/api/filter/ratelimit/RateLimitManagerTest.java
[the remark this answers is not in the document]
heathd: https://github.com/mockito/mockito/wiki/How-to-write-good-tests#dont-mock-value-objects

src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
```diff
@@ -62,12 +62,12 @@ private void setTooManyRequestsError() throws IOException {
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
-        final String authorization = requestContext.getHeaderString("Authorization");
         final String method = requestContext.getMethod();
```
kbottla: We could remove this and use `key.getMethod()` in the log statement below

src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
```diff
@@ -0,0 +1,49 @@
+package uk.gov.pay.api.filter;
+
+import org.apache.commons.lang3.StringUtils;
```
kbottla: Unused imports

src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.ws.rs.container.ContainerRequestContext;
+import java.util.Optional;
+
+public class RateLimiterKey {
+    
+    private String method;
+    private String key;
+    private String keyType;
+
+    public RateLimiterKey(String key, String keyType, String method) {
```
kbottla: Could also have path for logging purposes

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimitManager.java
```diff
@@ -0,0 +1,29 @@
...
+   
+    private RateLimiterConfig configuration;
+
+    public RateLimitManager(RateLimiterConfig config) {
+        configuration = config;
+    }
+
+    public int getAllowedNumberOfRequests(RateLimiterKey rateLimiterKey, String account) {
+        if(configuration.getElevatedAccounts().contains(account)) {
+            if("POST".equals(rateLimiterKey.getMethod())) {
```
kbottla: `HttpMethod.POST` ?

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -14,37 +15,38 @@
...
             throw new RedisException();
         }
 
-        if (count != null && count > getNoOfReqForMethod(method)) {
-            LOGGER.info(String.format("RedisRateLimiter - Rate limit exceeded for account [%s] and method [%s] - count: %d, rate allowed: %d", accountId, method, count, getNoOfReqForMethod(method)));
-            throw new RateLimitException();
+        if (count != null) {
+            int allowedNumberOfRequests = rateLimitManager.getAllowedNumberOfRequests(key, accountId);
+            if (count > allowedNumberOfRequests) {
+                LOGGER.info(String.format("RedisRateLimiter - Rate limit exceeded for account [%s] and type [%s] - count: %d, rate allowed: %d", accountId, key.getKeyType(), count, allowedNumberOfRequests));
```
kbottla: Would be useful to log path too (capture/payments..). Could come in handy when we move to structured logging

kbottla: I would prefer `method / http method` over `type`

src/main/java/uk/gov/pay/api/filter/ratelimit/RateLimiter.java
```diff
@@ -17,13 +18,13 @@ public RateLimiter(LocalRateLimiter localRateLimiter, RedisRateLimiter redisRate
...
-    public void checkRateOf(String accountId, String key, String method) throws RateLimitException {
+    public void checkRateOf(String accountId, RateLimiterKey key) throws RateLimitException {
         try {
-            redisRateLimiter.checkRateOf(accountId, key, method);
+            redisRateLimiter.checkRateOf(accountId, key);
         } catch (RedisException e) {
             LOGGER.warn("Exception occurred checking rate limits using RedisRateLimiter, falling back to LocalRateLimiter");
 
-            localRateLimiter.checkRateOf(accountId, key, method);
+            localRateLimiter.checkRateOf(accountId, key.getKey(), key.getMethod());
```
kbottla: `key.getMethod()` can be removed and can pass `RateLimiterKey`

src/main/java/uk/gov/pay/api/filter/RateLimiterKey.java
kbottla: @mrlumbu We have used API key before as account id was not available and missed on hashing API Key. Now that it is available we can use it for key.

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
jankowiakmaria: it's not always `POST` sometimes it's `POST-create_payment`, so maybe method is fine.

jankowiakmaria: it's not always `POST` sometimes it's `POST-create_payment`, so maybe method is fine.

PR #702, opened 2019-09-18
PP-5632: rate-limiting follow-up changes

jankowiakmaria:
Addressing review comments (https://github.com/alphagov/pay-publicapi/pull/701 and https://github.com/alphagov/pay-publicapi/pull/697):
* filtering empty values in StringToListConverter
* using method from rateLimiterKey in RateLimiterFilter
* removing unused imports from RateLimiterKey
* using HttpMethod.POST instead of string in RateLimitManager
* using removeEnd to remove trailing slash in PathHelper
* refactor LocalRateLimiter to use RateLimiterKey

### DIFF
--- a/src/main/java/uk/gov/pay/api/app/config/RateLimiterConfig.java
+++ b/src/main/java/uk/gov/pay/api/app/config/RateLimiterConfig.java
@@ -3,6 +3,7 @@ package uk.gov.pay.api.app.config;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import io.dropwizard.Configuration;
 
+import javax.validation.Valid;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 import java.util.Collections;
@@ -29,6 +30,7 @@ public class RateLimiterConfig extends Configuration {
     @Min(1)
     private int noOfReqForPostPerNode;
 
+    @Valid
     @JsonDeserialize(converter = StringToListConverter.class)
     private List<String> elevatedAccounts;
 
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/pay/api/app/config/RateLimiterConfig.java
```diff
@@ -27,6 +28,7 @@
     @Min(1)
     private int noOfReqForPostPerNode;
 
+    @Valid
```
kbottla: We need more than this. See https://github.com/alphagov/pay-connector/blob/86317f2faea6c1024d61fc4f7e3005f63d50e2fe/src/main/java/uk/gov/pay/connector/app/SqsConfig.java#L9 for example

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -44,7 +44,8 @@ synchronized void checkRateOf(String accountId, RateLimiterKey key)
         if (count != null) {
             int allowedNumberOfRequests = rateLimitManager.getAllowedNumberOfRequests(key, accountId);
             if (count > allowedNumberOfRequests) {
-                LOGGER.info(String.format("RedisRateLimiter - Rate limit exceeded for account [%s] and type [%s] - count: %d, rate allowed: %d", accountId, key.getKeyType(), count, allowedNumberOfRequests));
+                LOGGER.info(String.format("RedisRateLimiter - Rate limit exceeded for account [%s] and method [%s] - count: %d, rate allowed: %d",
```
kbottla: Unused import `javax.ws.rs.HttpMethod;`

jankowiakmaria: removed

src/main/java/uk/gov/pay/api/app/config/RateLimiterConfig.java
jankowiakmaria: Not needed. Each `noOf...` has `@Min(1)` validation which will cause the app to fail during the startup if the value is not provided (or 0).

PR #703, opened 2019-09-18
PP-5637 Search ledger by status version `1`

kbottla:
## WHAT
- PublicAPI uses different statuses to selfservice. With status_version=1, ledger returns the status on par with connector

### DIFF
--- a/src/main/java/uk/gov/pay/api/ledger/service/LedgerUriGenerator.java
+++ b/src/main/java/uk/gov/pay/api/ledger/service/LedgerUriGenerator.java
@@ -18,6 +18,7 @@ public class LedgerUriGenerator {
     }
 
     public String transactionsURIWithParams(Map<String, String> queryParams) {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/api/ledger/service/LedgerUriGenerator.java
```diff
@@ -18,6 +18,7 @@ public LedgerUriGenerator(PublicApiConfig configuration) {
     }
 
     public String transactionsURIWithParams(Map<String, String> queryParams) {
+        queryParams.put("status_version", "1");
```
heathd: how about documenting the meaning of this using an enum?

kbottla: Could potentially be candidate for ledger docs ? (for which we don't have API docs at the moment)

PR #704, opened 2019-09-18
PP-5648 make expiry date an optional field

SandorArpa:
## WHAT YOU DID
- as pointed out by our integrating partner, card_expiry field is not present when the transaction failed. We're removing the @NotNull annotation from CreateTelephonePaymentRequest
- update relevant tests

### DIFF
--- a/src/test/java/uk/gov/pay/api/it/telephone/CreateTelephonePaymentIT.java
+++ b/src/test/java/uk/gov/pay/api/it/telephone/CreateTelephonePaymentIT.java
@@ -10,6 +10,7 @@ import uk.gov.pay.api.utils.mocks.ConnectorMockClient;
 
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/api/it/telephone/CreateTelephonePaymentIT.java
```diff
@@ -126,6 +126,9 @@ public void createTelephonePaymentWithRequiredFields() {
 
     @Test
     public void returnExistingTelephonePayment() {
+        requestBody.put("card_expiry", "01/08");
+        createTelephonePaymentRequest
+                .withCardExpiry("01/08");
```
nimalank7: Why are we doing this? This integration test doesn't use optional fields so no need to pass along card_expiry. The mock can be modified to not return cardExpiry.

PR #722, opened 2019-10-02
PP-5285 OpenAPI specs for refund endpoints

kbottla:
## WHAT YOU DID
Prep to convert swagger to openAPI v3.0

- Adds annotations for refund endpoints defining parameters required, examples for related schemas

## HOW TO TEST
- `mvn compile` should now generate swagger (with openapi v3.0.0) with correct definitions for refund endpoints including examples

PR #727, opened 2019-10-03
PP-5285 Open API specs for `DirectDebit` endpoints

kbottla:
## WHAT YOU DID
Follows on from https://github.com/alphagov/pay-publicapi/pull/722 and now includes definitions for DirectDebit endpoints (for openapi v3.0.0)

## HOW TO TEST
- `mvn compile` should now generate swagger (with openapi v3.0.0) with correct definitions for direct debit endpoints including examples

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CreateDirectDebitPaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/CreateDirectDebitPaymentRequest.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
+import io.swagger.v3.oas.annotations.media.Schema;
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/pay/api/model/CreateDirectDebitPaymentRequest.java
```diff
@@ -45,21 +47,25 @@ public CreateDirectDebitPaymentRequest() {
     }
     
     @ApiModelProperty(value = "amount in pence", required = true, allowableValues = "range[1, 10000000]", example = "12000")
+    @Schema(description = "amount in pence", required = true, example = "12000")
```
stephencdaly: should there be `minimum` and `maximum` properties on this to replace the `allowableValues`?

kbottla: Good catch , updated

stephencdaly:
One more, is it worth committing the generated swagger file?

kbottla:
Not right now, there are payment endpoints that are unfinished and mixing both maven plugins may cause issues.

PR #732, opened 2019-10-04
PP-5285 Open API specs for `Card payment` endpoints

kbottla:
## WHAT YOU DID
- Swagger definitions for Card payment endpoints (for openapi v3.0.0)

## HOW TO TEST
- `mvn compile` should now generate swagger (with openapi v3.0.0) with correct definitions for card payment endpoints including examples

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CreatePaymentResult.java
+++ b/src/main/java/uk/gov/pay/api/model/CreatePaymentResult.java
@@ -3,6 +3,7 @@ package uk.gov.pay.api.model;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
+import io.swagger.v3.oas.annotations.media.Schema;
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/pay/api/model/CreatePaymentResult.java
```diff
@@ -63,21 +73,25 @@
...
+    @Schema(name = "metadata")
     private ExternalMetadata metadata;
 
     @JsonProperty
     @ApiModelProperty(name = "email", example = "citizen@example.org", required = false)
+    @Schema(name = "email", example = "citizen@example.org")
     private String email;
 
     @JsonProperty(value = "card_details")
-    @ApiModelProperty(name = "card_details", required = false)
```
SandorArpa: Is it accidental or purposeful removal?

kbottla: Good spot, replaced with `@Schema instead of adding and updated

PR #735, opened 2019-10-07
PP-5285 Fix specs for OpenAPI

kbottla:
## WHAT YOU DID
- Updated openapi specs to match current version of swagger


kbottla:
@danworth yeah it doesn't look into validation annotations. Will watch out if there are any extensions or dropwizard specific plugins that support it.

PR #779, opened 2019-11-27
PP-5761: Add X-Request-Id header to requests

oswaldquek:
This is so we can trace requests from publicapi to calls to downstream services
such as connector in splunk.

Note PP-5761 only covers connector so i haven't done anything for calls to direct-debit-connector.

### DIFF
--- a/src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
+++ b/src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
@@ -13,6 +13,7 @@ import javax.ws.rs.client.Client;
 import javax.ws.rs.client.Invocation;
 import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedHashMap;
 import javax.ws.rs.core.MultivaluedMap;
 
 import static org.assertj.core.api.Java6Assertions.assertThat;
@@ -60,5 +61,23 @@ public class ExternalServiceClientTest {
 
         verify(builder).headers(builderArgumentCaptor.capture());
         assertThat(builderArgumentCaptor.getValue().getFirst("X-Request-Id")).isEqualTo("123-easy-as-abc");
+    }
+    
+    @Test
+    public void assert_empty_headers_in_post_request_when_mdc_is_empty() {
+        MDC.clear();
+        externalServiceClient.post("http://example123.com");
+
+        verify(builder).headers(builderArgumentCaptor.capture());
+        assertThat(builderArgumentCaptor.getAllValues()).containsExactly(new MultivaluedHashMap<>());
+    }
+
+    @Test
+    public void assert_empty_headers_in_get_request_when_mdc_is_empty() {
+        MDC.clear();
+        externalServiceClient.get("http://example123.com");
+
+        verify(builder).headers(builderArgumentCaptor.capture());
+        assertThat(builderArgumentCaptor.getAllValues()).containsExactly(new MultivaluedHashMap<>());
     }
 }
--- a/src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
+++ b/src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
@@ -1,0 +1,64 @@
+package uk.gov.pay.api.clients;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.slf4j.MDC;
+
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.Invocation;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+
+import static org.assertj.core.api.Java6Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ExternalServiceClientTest {
+
+    private ExternalServiceClient externalServiceClient;
+
+    @Mock
+    private Client client;
+    @Mock
+    private WebTarget webTarget;
+    @Mock
+    private Invocation.Builder builder;
+    @Captor
+    private ArgumentCaptor<MultivaluedMap<String, Object>> builderArgumentCaptor;
+
+    @Before
+    public void setup() {
+        externalServiceClient = new ExternalServiceClient(client);
+        when(client.target(anyString())).thenReturn(webTarget);
+        when(webTarget.request()).thenReturn(builder);
+        when(builder.headers(any(MultivaluedMap.class))).thenReturn(builder);
+        when(builder.accept(MediaType.APPLICATION_JSON)).thenReturn(builder);
+    }
+    
+    @Test
+    public void assert_x_request_id_in_request_header_for_get() {
+        MDC.put("x_request_id", "abc");
+        externalServiceClient.get("http://example123.com");
+        
+        verify(builder).headers(builderArgumentCaptor.capture());
+        assertThat(builderArgumentCaptor.getValue().getFirst("X-Request-Id")).isEqualTo("abc");
+    }
+    
+    @Test
+    public void assert_x_request_id_in_request_header_for_post() {
+        MDC.put("x_request_id", "123-easy-as-abc");
+        externalServiceClient.post("http://example123.com");
+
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/pay/api/clients/ConnectorClient.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedHashMap;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.core.Response;
+import java.util.Map;
+import java.util.Optional;
+
+public class ConnectorClient {
```
stephencdaly: would this client not be usable for all the apps public api talks to?

src/main/java/uk/gov/pay/api/clients/ConnectorClient.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final Client client;
+
+    @Inject
+    public ConnectorClient(Client client) {
+        this.client = client;
+    }
+
+    public Response get(String url) {
+        MultivaluedMap<String, Object> headers = getHeaders();
```
stephencdaly: would be good to be consistent with the `post` method as to whether we inline the `getHeaders` call or not

oswaldquek:
I realise I don't have a unit test - please hold off code review for now

src/main/java/uk/gov/pay/api/clients/ConnectorClient.java
```diff
@@ -0,0 +1,43 @@
...
+        return client.target(url).request().headers(headers).accept(MediaType.APPLICATION_JSON).get();
+    }
+
+    private MultivaluedMap<String, Object> getHeaders() {
+        return Optional.ofNullable(MDC.get("x_request_id"))
+                    .filter(s -> !s.isBlank())
+                    .map(s -> new MultivaluedHashMap<String, Object>(Map.of("X-Request-Id", s)))
+                    .orElse(new MultivaluedHashMap<>());
```
stephencdaly: not saying change it, but is it really necessary to do the streaming stuff here? I guess it comes down to personal preference

oswaldquek: actually I'd prefer it with the streaming stuff as `return new MultivaluedHashMap<>(Map.of("X-Request-Id", MDC.get("x_request_id")));` will throw an exception if `MDC.get("x_request_id")` returns null.

stephencdaly: I more meant just doing a standard if statement checking whether the result is blank.

src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+    
+    @Test
+    public void assert_x_request_id_in_request_header_for_post() {
+        MDC.put("x_request_id", "123-easy-as-abc");
+        externalServiceClient.post("http://example123.com");
+
+        verify(builder).headers(builderArgumentCaptor.capture());
+        assertThat(builderArgumentCaptor.getValue().getFirst("X-Request-Id")).isEqualTo("123-easy-as-abc");
+    }
```
stephencdaly: could add tests to check the correct thing happens when there's no MDC?

src/main/java/uk/gov/pay/api/clients/ConnectorClient.java
oswaldquek: I don't think we should get into personal preferences... it's works for the functionality we're trying to achieve.

src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
```diff
@@ -0,0 +1,64 @@
...
+    public void setUp() {
+        externalServiceClient = new ExternalServiceClient(client);
+        when(client.target(anyString())).thenReturn(webTarget);
+        when(webTarget.request()).thenReturn(builder);
+        when(builder.headers(any(MultivaluedMap.class))).thenReturn(builder);
+        when(builder.accept(MediaType.APPLICATION_JSON)).thenReturn(builder);
+    }
+    
+    @Test
+    public void assert_x_request_id_in_request_header_for_get() {
```
stephencdaly: The naming of this test doesn't fit with our conventions

src/main/java/uk/gov/pay/api/clients/ConnectorClient.java
stephencdaly: yeh sure

src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
oswaldquek: I'm not we have coding standards for these anywhere ;) Alex doesn't appear to have minded when I did the same thing at https://github.com/alphagov/pay-connector/pull/1907/files.
I do think underscores are much more readable than camel case. 
Philosophically I think I'm paid to write readable code that works, rather than blindly follow conventions :)

stephencdaly: You make a fair point.
My counter argument would be that it's tidier to follow the unwritten conventions as it doesn't create the mental overhead of deciding which style to use. And if you come to expect things are a certain way it is a bit jarring to see it different in some places.
It's not a hill I'm going to die on though, so won't withhold approval over this.

oswaldquek: There's an existing proliferation of underscores in https://github.com/alphagov/pay-publicapi/blob/master/src/test/java/uk/gov/pay/api/validation/PaymentSearchValidatorTest.java too. I reckon people realise camel case becomes unreadable after a while.

src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
oswaldquek: I've added an extra test now.

src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
alexbishop1: Someone mentioned my name!

In general, all things being equal, I’ve become convinced over the years that `snake_case_is_easier_to_read` than `camelCaseIsToRead`.

However, in Java, `camelCaseIsKing` (`EXCEPT_FOR_CONSTANT_NAMES`) and there is a very, very strong convention that method names are camel case. This convention is so strong that I think it should probably considered an implied default in any Java coding standard unless there is guidance to the contrary.

As it happens, we do use underscores in some test method names and have done for longer than I’ve been here. An example would be `makeAppleStrudel_withIcingSugarOnTop_whenGuestsArrive()`. Here, the method name is actually a Frankenstein combination of camel case and snake case, with underscores used to split up the different aspects of the test: the expected result, a property of the result which perhaps differentiates this test from similar tests, and the action that causes the result. I’m personally opposed to using underscores in this way because it conflicts with general Java conventions. But other people aren’t, so our [PMD configuration allows underscores in JUnit test method names](https://github.com/alphagov/pay-publicapi/blob/master/ruleset.xml#L94-L95).

I haven’t seen full snake case used for tests before (I missed it in the other review, sorry!). While it is readable, I’m not generally in favour of this because it’s very un-Java-like. I don’t think there would be a huge readability difference if the test method name were `assertXRequestIdInRequestHeaderForGet()`.

This feels like a good thing to bring up at the next tech team meeting to see if we can find a consensus among the team.

oswaldquek: >This feels like a good thing to bring up at the next tech team meeting to see if we can find a consensus among the team.

Why not we do that and not delay this PR being merged. Of more importance is that we can start tracing requests properly in production.

src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
```diff
@@ -0,0 +1,83 @@
...
+        assertThat(builderArgumentCaptor.getValue().getFirst("X-Request-Id")).isEqualTo("abc");
+    }
+    
+    @Test
+    public void assert_x_request_id_in_request_header_for_post() {
+        MDC.put("x_request_id", "123-easy-as-abc");
+        externalServiceClient.post("http://example123.com");
+
+        verify(builder).headers(builderArgumentCaptor.capture());
+        assertThat(builderArgumentCaptor.getValue().getFirst("X-Request-Id")).isEqualTo("123-easy-as-abc");
```
stephencdaly: isn't it abc easy as 123?

oswaldquek: haha yup

src/test/java/uk/gov/pay/api/clients/ExternalServiceClientTest.java
alexbishop1: I never meant to suggest that we should delay the merging of this PR.

PR #802, opened 2019-12-10
PP-5893-Add-MDC-Logging-Filter

nimalank7:
Description:
- Renames method to add paymentId, mandateId and refundId to MDC context
- Adds helper method for this


stephencdaly:
Do we need to add a response filter to remove the MDC values, or is this handled already?

nimalank7:
This is handled already - see ClearMdcValuesFilter

PR #809, opened 2019-12-17
bau: Amend rate limit comment

oswaldquek:


src/main/resources/config/config.yaml
```diff
@@ -40,7 +40,7 @@ graphitePort: ${METRICS_PORT:-8092}
 jerseyClientConfig:
   disabledSecureConnection: ${DISABLE_INTERNAL_HTTPS:-false}
 
-rateLimiter:  # rate = noOfReq per perMillis
+rateLimiter:  # rate = noOfReq per second
```
jankowiakmaria: the original comment refers to the variable (from line 51) so maybe it would be better to just indicate that better.
The value is configurable and each environment can have it set up differently, so the new comment is not always correct.

oswaldquek: hmm yeah i think we should get rid of `perMillis` and set it to default of 1 second... it's kind of confusing. what do you think?

jankowiakmaria: I don't know. [This](https://github.com/alphagov/pay-publicapi#rate-limiting) sounds quite clear to me.
It doesn't seem we use custom configuration for any of our environments, but removing it would require quite a lot code change. I'm not in favour of that.

PR #824, opened 2020-01-09
BAU: code clean up

jankowiakmaria:
Changes:
* replace deprecated methods
* use Charset constant (instead of string)
* extract common parts from if in ConnectorMockClient
* replace switch with if (too few branches) - PaymentWithAllLinks
* invert boolean method - the usages are always in negated context
* remove redundant 'throws' clause
* remove unnecessary import(s)
* inline variables

### DIFF
--- a/src/main/java/uk/gov/pay/api/validation/MaxLengthValidator.java
+++ b/src/main/java/uk/gov/pay/api/validation/MaxLengthValidator.java
@@ -3,7 +3,7 @@ package uk.gov.pay.api.validation;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 class MaxLengthValidator {
-    static boolean isValid(String value, int maxLength) {
-        return isBlank(value) || value.length() <= maxLength;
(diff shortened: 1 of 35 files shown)


src/main/java/uk/gov/pay/api/validation/MaxLengthValidator.java
```diff
@@ -3,7 +3,7 @@
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 class MaxLengthValidator {
-    static boolean isValid(String value, int maxLength) {
-        return isBlank(value) || value.length() <= maxLength;
+    static boolean isInvalid(String value, int maxLength) {
+        return !isBlank(value) && value.length() > maxLength;
```
kbottla:
```suggestion
        return isNotBlank(value) && value.length() > maxLength;
```
might make it more readable

jankowiakmaria: Makes sense - I let the IDE do the job.
I'll fix it in the follow-up PR.

PR #826, opened 2020-01-10
PP-6001 Upgrade dropwizard

kbottla:
## WHAT 
- Upgrades dropwizard to v2.0.0
- Added `dropwizard-dependencies` as parent as v2.0.0 upgrade doesn't specify transitive dependency versions anymore https://www.dropwizard.io/en/release-2.0.x/manual/upgrade-notes/upgrade-notes-2_0_x.html

  We can also include BOM for transitive dependency versions but versions can't be overridden and we may end up with multiple revisions of same library. For example, `jackson-version` current used is `2.10.2` but dropwizard uses `2.10.1`

- Upgraded dropwizard-sentry library to latest version as container might fail to start with error `java.lang.NoSuchMethodError: 'com.google.common.collect.ImmutableList org.dhatim.dropwizard.sentry.logging.SentryAppenderFactory.getFilterFactories()

- Upgraded pay-java-commons library which is on dropwizard v2.0.0 and defines required library `org.glassfish.jersey.bundles.repackaged`

- Register `ReturnUrlValidator` instance (created by Guice) explicitly as Jersey is not able to create an instance of ReturnUrlValidator and throws following RuntimeException

> "org.glassfish.hk2.api.MultiException: A MultiException has 1 exceptions.  They are:\n1. org.glassfish.hk2.api.UnsatisfiedDependencyException: There was no object available for injection at SystemInjecteeImpl(requiredType=URLValidator,parent=ReturnUrlValidator,qualifiers={},position=0,optional=false,self=false,unqualified=null,807417419)\n\n\tat org.jvnet.hk2.internal.ThreeThirtyResolver.resolve(ThreeThirtyResolver.java:51)\n\tat 

- Few fixes for new version and validators annotation class

### DIFF
--- a/src/main/java/uk/gov/pay/api/app/config/PublicApiConfig.java
+++ b/src/main/java/uk/gov/pay/api/app/config/PublicApiConfig.java
@@ -2,7 +2,7 @@ package uk.gov.pay.api.app.config;
 
 import com.bendb.dropwizard.redis.JedisFactory;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.cache.CacheBuilderSpec;
+import com.github.benmanes.caffeine.cache.CaffeineSpec;
 import io.dropwizard.Configuration;
 
 import javax.validation.Valid;
@@ -36,7 +36,7 @@ public class PublicApiConfig extends Configuration {
     private String apiKeyHmacSecret;
 
     @NotNull
-    private CacheBuilderSpec authenticationCachePolicy;
+    private CaffeineSpec authenticationCachePolicy;
 
     @Valid
     @NotNull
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/api/app/config/PublicApiConfig.java
```diff
@@ -2,7 +2,7 @@
 
 import com.bendb.dropwizard.redis.JedisFactory;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.cache.CacheBuilderSpec;
+import com.github.benmanes.caffeine.cache.CaffeineSpec;
```
alexbishop1: What’s this change for?

kbottla: Dropwizard uses `CaffeineSpec` instead of `CacheBuilderSpec` in v2.0.0 for `CachingAuthenticator`

https://github.com/dropwizard/dropwizard/blob/master/dropwizard-auth/src/main/java/io/dropwizard/auth/CachingAuthenticator.java#L42

PR #833, opened 2020-01-21
PP-6010 Add test for not finding charge for nonexisting reference

nimalank7:
Description:
- Moves SearchPaymentsIT.shouldNotFindAnyChargeByANonExistingReference from pay-end-to-end 'publicapi' to PaymentResourceSearchIT


src/test/java/uk/gov/pay/api/it/PaymentResourceSearchIT.java
```diff
@@ -17,6 +17,7 @@
 
 import java.io.InputStream;
 import java.time.ZonedDateTime;
+import java.util.ArrayList;
```
stephencdaly: looks like this import isn't needed

nimalank7: Done

PR #843, opened 2020-01-31
bau: Collapse create payment pact test

oswaldquek:
Pact tests should test the API, rather than test individual bits of
functionality such as language, metadata, delayed capture and prefilled
cardholder details. Variants of these should be integration tests in the
connector repo.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation


stephencdaly:
Seems sensible, I saw this the other day and thought it was wrong. Is it worth having one more for creating a payment providing the minimum fields? Is it part of the contract we're testing that certain fields are optional?

oswaldquek:
@stephencdaly yeah sounds sensible to have one with minimum fields. I'll add a test.

PR #845, opened 2020-02-03
Pp 6077 moto disabled

oswaldquek:
## WHAT YOU DID
Pact test for creating a moto payment when the gateway account is not configured with moto enabled.

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation


stephencdaly:
Only thing I'd question is putting the moto fields and methods highest up in various classes. This won't have any impact on outputs so is more of a style thing. Just find it slightly jarring.

oswaldquek:
@stephencdaly fixed

PR #853, opened 2020-02-05
PP-6077 - Modify Pact for get payment moto field

nimalank7:
Description:
- Adds moto field to get payment pact tests. No real need to add a get payment with moto as a previous pact test checks whether a moto payment can be created


stephencdaly:
While you're here, could we merge most of these pact tests down into a single one for fields that are always returned by connector, like what was done for the create payment pacts in https://github.com/alphagov/pay-publicapi/pull/843?

oswaldquek:
@stephencdaly yeah I was thinking that. @nimalank7 are you up for the challenge? 😃

nimalank7:
Sure happy to do so! Can I do that in a separate PR? That way we can move the ticket to review?

PR #862, opened 2020-02-07
PP-6077 - Refactor get payment with get payment fee

nimalank7:
Description:
- Collapses get-payment-with-fee into get-payment as Pact should test the API not individual bits of functionality


src/test/resources/pacts/publicapi-connector-get-payment.json
```diff
@@ -35,6 +35,8 @@
             "some_key": "key"
           },
           "amount": 100,
+          "fee": 5,
+          "net_amount": 345,
```
oswaldquek: i'm not sure about this. isn't net_amount = amount - fee ?

nimalank7: I ran the pact locally with this and it failed telling me it expected it at 345. This is because we added corporate surcharge which gives it to 350.

oswaldquek: ah right, ok

PR #886, opened 2020-03-02
PP-5955: Add description to moto payments flag

oswaldquek:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/CardPayment.java
+++ b/src/main/java/uk/gov/pay/api/model/CardPayment.java
@@ -36,7 +36,6 @@ public class CardPayment extends Payment {
     private final SupportedLanguage language;
 
     @JsonProperty("delayed_capture")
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/api/model/CardPayment.java
```diff
@@ -145,7 +144,9 @@ public SupportedLanguage getLanguage() {
     public boolean getDelayedCapture() {
         return delayedCapture;
     }
-    
+
+    @ApiModelProperty(value = "moto payment flag", example = "false")
+    @Schema(description = "moto payment flag", example = "false")
```
stephencdaly: Shouldn't it be "Mail Order / Telephone Order (MOTO) payment flag"?

PR #946, opened 2020-06-30
PP-6708 Upgrade to Junit5

kbottla:
## WHAT YOU DID
- Upgrades most of the unit tests to Junits.
  Parameterised tests are not updated in this change.
- Removed deprecated rule `Exception.none()` and replaced with `assertThrows` which is recommended

## How to test

- All unit tests still pass on CI


src/test/java/uk/gov/pay/api/filter/RateLimiterFilterTest.java
```diff
@@ -59,24 +60,23 @@ public void setup() {
...
-        try {
-            rateLimiterFilter.filter(mockContainerRequestContext);
-        } catch (WebApplicationException e) {
-            Response response = e.getResponse();
-            assertEquals(429, response.getStatus());
-            assertEquals("application/json", response.getHeaderString("Content-Type"));
-            assertEquals("utf-8", response.getHeaderString("Content-Encoding"));
-            assertEquals("{\"code\":\"P0900\",\"description\":\"Too many requests\"}", response.getEntity());
-        }
+        WebApplicationException webApplicationException = assertThrows(WebApplicationException.class,
```
alexbishop1, line 73: So we can assert an exception is thrown and then carry on doing more stuff without even needing a try/catch block? Neat!

src/test/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiterTest.java
```diff
@@ -77,29 +78,27 @@ public void rateLimiterSetTo_2CallsPerSecond_shouldAllow2ConsecutiveCallsWithSam
...
 
         when(jedis.incr(anyString())).thenReturn(1L, 2L, 3L);
 
-        try {
-            redisRateLimiter.checkRateOf(accountId, rateLimiterKey);
-            redisRateLimiter.checkRateOf(accountId, rateLimiterKey);
+        redisRateLimiter.checkRateOf(accountId, rateLimiterKey);
+        redisRateLimiter.checkRateOf(accountId, rateLimiterKey);
+
+        assertThrows(RateLimitException.class, () -> {
```
alexbishop1, line 95: I assume this test doesn’t actually throw a `RedisException` then?

PR #1001, opened 2020-09-15
PP-7154 Add dummy url for dd connector for PaaS

danworth:
We are no deploying DDConnector into PaaS but for now Publicapi
still requires the url. This is a dummy value so that Publicapi will
start. This can be removed as part of removing DD from Publicapi.

## WHAT YOU DID
We no longer deploy dd connector and do not provide it's url via the app-catalogue. I've tried this by manually pushing to staging and it works.


rorymalcolm (approved):
LGTM

PR #1004, opened 2020-09-18
PP-6343: Replace Jedis with lettuce-core library

oswaldquek:
Follows the basic usage according to https://github.com/lettuce-io/lettuce-core/wiki/Basic-usage

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
+++ b/src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
@@ -2,12 +2,11 @@ package uk.gov.pay.api.filter.ratelimit;
 
 import com.google.inject.Inject;
 import com.google.inject.OutOfScopeException;
-import io.lettuce.core.RedisClient;
-import io.lettuce.core.api.StatefulRedisConnection;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.api.app.config.RateLimiterConfig;
 import uk.gov.pay.api.filter.RateLimiterKey;
+import uk.gov.pay.api.managed.RedisClientManager;
 
 import javax.inject.Singleton;
 import java.time.LocalDateTime;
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -52,20 +53,14 @@ synchronized void checkRateOf(String accountId, RateLimiterKey key)
 
     synchronized private Long updateAllowance(String key) {
         String derivedKey = getKeyForWindow(key);
+        StatefulRedisConnection<String, String> connection = redisClient.connect();
```
kbottla, line 56: Do we need to close this connection or does lettuce takes care of it ? Java docs says `connect()` opens new connection to Redis https://lettuce.io/core/release/api/io/lettuce/core/RedisClient.html#connect-- 

Also with Jedis, we get connection pool. Is this something that will be useful ?

kbottla (commented):
Looks good to me. Just a comment about managing connection (ie., if it is to be closed explicitly or management comes with lettuce by default) and if a connection pool is useful if we are opening connection for every request

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
oswaldquek: Thanks for catching this out. I think the `StatefulRedisConnection` can be a field in the class. However, given that `updateAllowance` is a synchronized method, I don't think there's any value in using a connection pool here.

oswaldquek: @kbottla I've closed the connection via the (renamed) RedisClientManager now.

kbottla (commented):
Looks good to me.  
A minor comment about using wrapper for RedisClient to keep connection management from RedisRateLimiter, but is not blocking..

Would be good to get more eyes on ADR and if we want to plan further steps to make Redis mandatory in all environments.

src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -96,4 +99,9 @@ private String getKeyForWindow(String key) throws OutOfScopeException {
         return key + window;
     }
 
+    public void closeRedisConnection() {
```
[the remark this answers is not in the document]
oswaldquek: Good shout. I've now managed the RedisClient connection from within the RedisClientManager instead.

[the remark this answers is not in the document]
oswaldquek: I felt it was a little over the top to create a wrapper class.

kbottla (approved):
LGTM!

PR #1006, opened 2020-09-22
PP-6343: ADR for using the lettuce-core library instead of dropwizard…

oswaldquek:
…-redis

### DIFF
--- a/docs/arch/adr-001-use-lettuce-core-lib-instead-of-dropwizard-redis.md
+++ b/docs/arch/adr-001-use-lettuce-core-lib-instead-of-dropwizard-redis.md
@@ -2,32 +2,41 @@
 
 ## Context
 
-There was an [intention](https://payments-platform.atlassian.net/browse/PP-6343) to use the 
-[dropwizard-redis](https://github.com/dropwizard/dropwizard-redis) library. The advantages of using this library are:
+We were using an [unsupported dropwizard-redis](https://github.com/benjamin-bader/droptools/tree/master/dropwizard-redis) bundle. Dropwizard now offer a managed bundle which provides a managed redis. This bundle uses `lettuce-core` instead of `jedis`.
+
+[We wanted to upgrade to this new library](https://payments-platform.atlassian.net/browse/PP-6343).
+
+The advantages of using this library are:
 
 * Configuration
 * Client lifecycle management
 * Client health checks
 * Dropwizard Metrics integration
 
-The dropwizard-redis library's io.dropwizard.redis.RedisClientFactory class 
+However, the dropwizard-redis library's io.dropwizard.redis.RedisClientFactory class 
 [expects](https://github.com/dropwizard/dropwizard-redis/blob/master/src/main/java/io/dropwizard/redis/RedisClientFactory.java#L54) 
-tries to make a connection with Redis on application startup. If a connection cannot be made an exception will be thrown 
-which causes the application to fail to start up. This has some unintended consequences:
+to make a connection with Redis on application startup. If a connection cannot be made an exception will be thrown 
+which causes the application to fail to start up. 
+
+This has some unintended consequences:
 
 * Mandating a connection on startup causes many integration tests to fail. Making a redis connection available for every relevant test is a fairly big change.
 * Mandating a connection might affect running Publicapi in dev/local environments.
 
+In order to work around this issue we could either:
+
+1. adapt the dropwizard-redis library so that it doesn't crash if the redis connection is not available
+2. use lettuce-core directly
+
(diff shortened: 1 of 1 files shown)


docs/arch/adr-001-use-lettuce-core-lib-instead-of-dropwizard-redis.md
```diff
@@ -0,0 +1,34 @@
...
+There was an [intention](https://payments-platform.atlassian.net/browse/PP-6343) to use the 
+[dropwizard-redis](https://github.com/dropwizard/dropwizard-redis) library. The advantages of using this library are:
+
+* Configuration
+* Client lifecycle management
+* Client health checks
+* Dropwizard Metrics integration
+
+The dropwizard-redis library's io.dropwizard.redis.RedisClientFactory class 
+[expects](https://github.com/dropwizard/dropwizard-redis/blob/master/src/main/java/io/dropwizard/redis/RedisClientFactory.java#L54)
```
heathd, line 14: it's not evident/apparent to me why this line of code indicates that Redis is required to be running at application startup. The link shows that `RedisClientFactory.build()` attempts to open a connection.

Digging further through the code I can see that the `RedisClientBundle` [builds the connection](https://github.com/dropwizard/dropwizard-redis/blob/master/src/main/java/io/dropwizard/redis/RedisClientBundle.java#L29) on startup.

Wouldn't a cleaner/simpler solution to be to override the `run` method of `RedisClientBundle`  to make it catch connection exceptions during startup and switch to a 'disabled/not available' mode? That way we can continue to use the maintained and instrumented official dropwizard client.

docs/arch/adr-001-use-lettuce-core-lib-instead-of-dropwizard-redis.md
```diff
@@ -0,0 +1,34 @@
...
+* Mandating a connection on startup causes many integration tests to fail. Making a redis connection available for every relevant test is a fairly big change.
+* Mandating a connection might affect running Publicapi in dev/local environments.
+
+## Decision
+
+We will use the [lettuce-core](https://github.com/lettuce-io/lettuce-core) instead. Reasons for this are:
+
+* The dropwizard-library uses the lettuce-core libarary internally anyway
+* Publicapi is currently designed to be able to work without Redis
+* Redis in our environments is currently a single instance
```
heathd, line 26: Why is this a reason to use `lettuce-core`?

heathd:
I've not dug too deep into the code and I'm not going to insist on changing tack now that you've implemented something, but I think the reasoning needs to be set out a bit more clearly here.

docs/arch/adr-001-use-lettuce-core-lib-instead-of-dropwizard-redis.md
oswaldquek: Yes I should've made it clearer in the ADR but the line of code (`redisClient.connect(codec)`) will throw an exception if it can't connect to Redis.

oswaldquek: I think overriding the `run` method and switching to a 'disabled/not available' mode makes things unnecessarily complicated because we'd still need to manage the `RedisClientManager` (see https://github.com/dropwizard/dropwizard-redis/blob/master/src/main/java/io/dropwizard/redis/RedisClientFactory.java#L57) but we can't manage it if we don't have a `connection`, and we can't have a `connection` if a connection to Redis can't be made. Managing the lifecycle  is part of the point of using the dropwziard-redis library. 
As a workaround we've already built a custom RedisClientManager (see https://github.com/alphagov/pay-publicapi/blob/master/src/main/java/uk/gov/pay/api/managed/RedisClientManager.java) which doesn't require a connection and this code isn't much extra effort.

docs/arch/adr-001-use-lettuce-core-lib-instead-of-dropwizard-redis.md
oswaldquek: Yeah this isn't a valid reason. I originally thought this was a reason because even with a failover HA will failover to another instance but it would mean the data is wiped clean. But I think that's not correct.

docs/arch/adr-001-use-lettuce-core-lib-instead-of-dropwizard-redis.md
oswaldquek: I've updated the point to say `redisClient.connect(codec)` will throw an exception if it can't connect to Redis and thus cause the application to fail to start up.

heathd (approved):

PR #1007, opened 2020-09-22
PP-6670 add settlement summary to refunds

SandorArpa:
## WHAT YOU DID
- add settlement_summary to refund responses
- settlement_summary is the same as payments' one but will never have capture details. refunds from connector are always an empty object {}
- add/extend tests

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/search/card/RefundForSearchRefundsResult.java
+++ b/src/main/java/uk/gov/pay/api/model/search/card/RefundForSearchRefundsResult.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 import io.swagger.v3.oas.annotations.media.Schema;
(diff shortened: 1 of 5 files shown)


stephencdaly (changes requested):
Looks good, just a few minor things with the swagger annotations.

src/main/java/uk/gov/pay/api/model/search/card/RefundForSearchRefundsResult.java
```diff
@@ -95,6 +104,13 @@ public RefundLinksForSearch getLinks() {
         return links;
     }
 
+    @ApiModelProperty(hidden = true, dataType = "uk.gov.pay.api.model.SettlementSummary")
+    @JsonProperty("settlement_summary")
+    @Schema(hidden = true)
```
stephencdaly, line 109: I think the annotations on either the getter or the field can be removed?

SandorArpa: I agree, it is strange way but I just followed the existing way of annotating fields/getters

stephencdaly (approved):
LGTM

PR #1008, opened 2020-09-23
PP-6670 add a pact test containing settled date

SandorArpa:
## WHAT YOU DID
- add a pact test that checks for settled date


kbottla (approved):
LGTM!

PR #1024, opened 2020-10-12
PP-7256 refactor refund use its own settlement summary class

SandorArpa:
## WHAT YOU DID
- this change is required so our swagger tech docs are less confusing for the end user. although it makes a bit of code duplication, it also removes unnecessary overhead for the user and makes it clear that refunds' settlement summary contains settlement date only.


stephencdaly (approved):
Looks good to me!

PR #1042, opened 2020-11-09
PP-7104 Log when API keys are used

stephencdaly:
Log at INFO level when a valid API key is used. Log the token_link for the API key, which is returned by public auth.

Log at WARN level when an API key with an invalid checksum is used. Log the IP address the request originates from to help us detect brute force attempts.

Log at WARN level when an invalid API key with a valid checksum is used.

Log at WARN level when a revoked API key is used. Log the token_link for the API key, which is returned by public auth.

Remove logic that maps a non-existent/unknown value for `token_type` returned from public auth to TokenPaymentType.CARD. Public auth will always return a `token_type`, and if it doesn't this will result in a `com.fasterxml.jackson.databind.exc.InvalidFormatException` which will be wrapped in a 500 response. This will show up in Sentry so we should be able to catch it if there are any future issues introduced.

Add an integration test to ensure this is the case

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/api/auth/AccountAuthenticator.java
+++ b/src/main/java/uk/gov/pay/api/auth/AccountAuthenticator.java
@@ -5,7 +5,8 @@ import io.dropwizard.auth.Authenticator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.api.app.config.PublicApiConfig;
-import uk.gov.pay.api.model.TokenPaymentType;
+import uk.gov.pay.api.model.publicauth.AuthResponse;
+import uk.gov.pay.commons.model.ErrorIdentifier;
 
 import javax.inject.Inject;
 import javax.ws.rs.ServiceUnavailableException;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/api/model/publicauth/AuthResponse.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.pay.api.model.publicauth;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import uk.gov.pay.api.model.TokenPaymentType;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
```
sfount, line 9: 👀 👍

src/main/java/uk/gov/pay/api/auth/AccountAuthenticator.java
```diff
@@ -41,8 +42,19 @@ public AccountAuthenticator(Client client, PublicApiConfig configuration) {
                 .get();
 
         if (response.getStatus() == OK.getStatusCode()) {
-            return readAccountFromResponse(response);
+            AuthResponse authResponse = response.readEntity(AuthResponse.class);
```
sfount, line 45: Organising this into a simple class is great 👍 . It looks like the behaviour to assign the default `CARD` token type if there is none present is removed here, just sense checking that that will play nice with serialisation/ deserialisation here and upstream (public auth).

src/main/java/uk/gov/pay/api/filter/AuthorizationValidationFilter.java
```diff
@@ -31,38 +36,45 @@ public AuthorizationValidationFilter(PublicApiConfig configuration) {
...
     }
 
     @Override
-    public void destroy() {}
+    public void destroy() {
+    }
 
-    private boolean isValidAuthorizationHeader(String authorization) {
-        return authorization!= null
+    private boolean isValidAuthorizationHeader(String authorization, ServletRequest request) {
```
sfount, line 58: If we're only going to need the remote address is it worth just passing that through as a string? 

(After typing this I'm not so sure about it, the method signature would be confusing to understand as this method doesn't actually use the request but passes it on to the next. Happy for this to be done either way!)

sfount (commented):
This looks great! As discussed a question around the default behaviour when serialising the request body

sfount (approved):

PR #1044, opened 2020-11-10
PP-7104 Fix logging when API key fails checksum

stephencdaly:
Move the initialisation of filters for the application so that the request id is added to the MDC for logging before the AuthorizationValidationFilter runs.

Fix logging the "remote_address" for the line logged when the API key fails checksum validation so that it logs the client address obtained from the "X-Forwarded-For" header, which is added by the proxies the request passes through.


sfount (approved):
Really neat solution 👍

PR #1049, opened 2020-11-24
PP-7429 Add token_link and remote_address to MDC

kbottla:
## WHAT YOU DID
- Added `token_link` and `remote_address` to MDC so these are logged for all log lines.
  Changed the priority of RateLimiterFilter so this is invoked after LoggingMDCRequestFilter (which adds keys to MDC)

- Although keys are added to MDC, a few logs below won't have the same as logging happens before/during API authentication and before keys are added to MDC

    "POST to /v1/payments began","logger_name":"uk.gov.pay.logging.LoggingFilter" ...
    "[] - GET to http://publicauth/v1/api/auth began",  ....
    "[] - GET to http://publicauth/v1/api/auth ended - total time 147ms", ...

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/LoggingMDCRequestFilter.java
+++ b/src/main/java/uk/gov/pay/api/filter/LoggingMDCRequestFilter.java
@@ -1,7 +1,7 @@
 package uk.gov.pay.api.filter;
 
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.MDC;
+import uk.gov.pay.api.auth.Account;
 
 import javax.annotation.Priority;
 import javax.ws.rs.Priorities;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/api/filter/LoggingMDCRequestFilter.java
```diff
@@ -11,18 +11,26 @@
...
 @Provider
 @Priority(Priorities.USER)
 public class LoggingMDCRequestFilter implements ContainerRequestFilter {
 
     @Override
     public void filter(ContainerRequestContext requestContext) {
-        MDC.put(GATEWAY_ACCOUNT_ID, getAccountId(requestContext));
+        Optional<Account> mayBeAccount = getAccount(requestContext);
+        MDC.put(GATEWAY_ACCOUNT_ID, mayBeAccount.map(Account::getName).orElse(EMPTY));
+        MDC.put("token_link", mayBeAccount.map(Account::getTokenLink).orElse(EMPTY));
```
sfount, line 29: as this is logged across both public api and public auth (and could be potentially useful over in self service logs), it's potentially a good candidate for a shared key?

sfount (approved):
Looks good to me 👍

PR #1062, opened 2020-12-04
PP-7526 Do not rate limit calls to healthcheck

danworth:
Do not rate limit calls to the healthcheck endpoint. This is used by the
loadbalancer to test if an instance is healthy and a 429 response is considered
unhealthy. This could lead to the loadbalancer incorrectly setting the instance
as unhealthy.

## WHAT YOU DID
Although we're not sure whether it caused or contributed to an incident where the Publicapi Loadbalancer began returning 503 responses it isn't ideal that we are rate limiting its requests to the `/healthcheck` endpoint.

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
+++ b/src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
@@ -62,6 +62,10 @@ public class RateLimiterFilter implements ContainerRequestFilter {
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
+        if ("healthcheck".equals(requestContext.getUriInfo().getPath())) {
+            return;
+        }
+
         String accountId = getAccountId(requestContext);
         RateLimiterKey key = RateLimiterKey.from(requestContext, accountId);
         try {
--- a/src/test/java/uk/gov/pay/api/filter/RateLimiterFilterTest.java
+++ b/src/test/java/uk/gov/pay/api/filter/RateLimiterFilterTest.java
@@ -20,9 +20,11 @@ import javax.ws.rs.core.UriInfo;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/api/filter/RateLimiterFilter.java
```diff
@@ -62,6 +62,10 @@ private void setTooManyRequestsError() throws IOException {
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
+        if ("healthcheck".equals(requestContext.getUriInfo().getPath())) {
```
danworth, line 65: is it being overly defensive to actually do a null check on the return value of `getUriInfo` before calling `getPath`? Every request to publicapi goes through this so it should be robust but would `UriInfo` every be `null`?
```
if (requestContext.getUriInfo() != null && "healthcheck".equals(requestContext.getUriInfo().getPath())) {
```

alexbishop1: I think `requestContext.getUriInfo()` should never return null. Looking at the methods on `[UriInfo](https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/UriInfo.html)` shows it has methods like `getAbsolutePath()` and there’s always an absolute path, so it makes sense there will always be a `UriInfo` to contain it.

Also, if we look at the source for the `ContainerRequestContext` in Glassfish, we can see `getUriInfo()` is defined like this:

```java
@Override
public ExtendedUriInfo getUriInfo() {
    return uriRoutingContext;
}
```

`uriRoutingContext` is a field initialised in the constructor:

```java
this.uriRoutingContext = new UriRoutingContext(this);
```

Nothing else in the class seems to mutate it and there’s other code in there that seems to assume it’s not null.

src/test/java/uk/gov/pay/api/filter/RateLimiterFilterTest.java
```diff
@@ -44,13 +47,13 @@ public void setup() {
 
         Account account = new Account(ACCOUNT_ID, TokenPaymentType.CARD, "some-token-link");
         SecurityContext mockSecurityContext = mock(SecurityContext.class);
-        when(mockSecurityContext.getUserPrincipal()).thenReturn(account);
+        lenient().when(mockSecurityContext.getUserPrincipal()).thenReturn(account);
```
alexbishop1, line 50: I’d prefer it we moved these so they’re only in methods where they’re actually used.

alexbishop1 (approved):

src/test/java/uk/gov/pay/api/filter/RateLimiterFilterTest.java
danworth: yeah ok I've updated 👍

alexbishop1 (approved):

PR #1066, opened 2020-12-14
PP-7560 Clear MDC at both the beginning and end of a request

alexbishop1:
We are observing in our logs that sometimes the remote_address from a previous request seems to be logged even though the MDC should have been cleared at the end of that request.

Additionally clear the MDC at the beginning of the request (it should have no values at this point so we shouldn’t lose anything we want) to see if this makes things better.

with @katstevens


katstevens (approved):

heathd:
will this run before the `RestClientLoggingFilter` as well?

heathd:
think it should do because that's being used to invoke PublicAuth during the Auth filter, so the clear should run before it

PR #1077, opened 2021-01-08
PP-7589 Refactor TelephonePaymentNotificationResource

stephencdaly:
I found the names of methods and interaction with the service slightly confusing. The reason it's like this is because public api will respond with a 200 or 201 depending on the status code from connector.

Refactored so that the service is responsible for constructing the public api response from the connector response, and returns a Pair of the response and the status code to reply with.


kbottla (approved):
LGTM!

PR #1102, opened 2021-03-02
PP-7834: Delete publicapi -> direct-debit-connector pacts

oswaldquek:
Deleting test cases where pacts between publicapi and direct-debit-connector
have been published to the pact broker. We need to delete these as there is no
verified pact between publicapi and direct-debit-connector with the new
"test-fargate" tag when running compatibility checks on the deploy-publicapi
(https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/apps-test-ecr/jobs/deploy-publicapi)
job on concourse.

src/test/java/uk/gov/pay/api/it/directdebit/pact/CancelPaymentIT was not
deleted here as it looks useful. It has been moved to
src/test/java/uk/gov/pay/api/it/CancelPaymentIT.

More direct debit related classes (like the services and resources) could be
deleted but that is out of scope for PP-7834.


stephencdaly (approved):
🗑️ 😢

PR #1111, opened 2021-03-30
PP-7896 Remove swagger maven plugin

kbottla:
## WHAT YOU DID
- Removed `swagger-maven-plugin` as we no longer intend to maintain swagger v2
- Removed swagger annotations used to generate `swagger.json` from resources

TODO: To remove annotations on model classes in a separate PR


stephencdaly (approved):
LGTM

PR #1112, opened 2021-03-30
PP-7896 Remove swagger (v2) annotations from models

kbottla:
## WHAT YOU DID
- Remove swagger annotations from models
- Removed `swagger-jersey2-jaxrs` library as it is no longer needed


SandorArpa (approved):
LOKTM

PR #1178, opened 2021-07-13
PP-8210 - Return 403 when PSP not linked with gateway account

nimalank7:
Description:
- Return 403 and error code of P0940 with the description message when we receive 400 from Connector with the error identifier as 'ACCOUNT_NOT_LINKED_WITH_PSP' and the message as 'Payment provider details are not configured on this account'
- Removed 'public' from PaymentErrorTest as methods don't need to be in JUnit 5
- Removed unused emocks in PaymentsResourceCreatePaymentTest


kbottla (approved):
LGTM!

PR #1185, opened 2021-07-20
PP-8210 - Pact test for gateway account credentials in created

nimalank7:
Description:
- I re-used the state as I only needed a gateway_account_credential in a CREATED state
- Checks if Connector returns 400 when the gateway_account_credential is in a CREATED state


stephencdaly (approved):
LGTM!

PR #1199, opened 2021-08-02
PP-8224 Add github actions to dependabot file

stephencdaly:


kbottla (approved):
LGTM!

From the offline discussion, it looks like dependabot updates github actions to latest commit sha if actions are pinned to specific commit.

PR #1230, opened 2021-08-31
PP-8550 Rate limit requests for low traffic accounts

kbottla:
## WHAT YOU DID
- Rate limits requests for low traffic accounts as per the configuration (RATE_LIMITER_LOW_TRAFFIC_ACCOUNTS and related env variables).
	- Number of post requests are defaulted to 1 request per minute and per bucket (capture, create and the rest)
	- Number of get requests allowed will be the same as the rate limit for standard accounts (4500 requests per minute which is equivalent to 75 req per second) but rate limit interval is per minute.

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
+++ b/src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
@@ -17,13 +17,11 @@ public class RedisRateLimiter {
     private static final Logger LOGGER = LoggerFactory.getLogger(RedisRateLimiter.class);
 
     private RateLimitManager rateLimitManager;
-    private final int perMillis;
     private RedisClientManager redisClientManager;
 
     @Inject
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
```diff
@@ -71,22 +70,25 @@ synchronized private Long updateAllowance(String key) {
      * Depends on perMillis
      * <p>
      * - perMillis < 1000 : Window considered for milliseconds
-     * - perMillis >=1000 && <=60000 : Window considered for seconds(s)
+     * - perMillis >=1000 && <60000 : Window considered for seconds(s)
+     * - perMillis >=1000 && <=60000 : Window considered for minute(s)
```
stephencdaly, line 74: should this comment be 
```   * - perMillis >=60000 && <3600000 : Window considered for minute(s)```

stephencdaly (approved):
LGTM just think there's a comment that's wrong

stephencdaly (approved):

PR #1243, opened 2021-09-15
PP-8535 add new and update pact tests with 3ds information

SandorArpa:
## WHAT YOU DID
- update connector get one payment with authorisation_summary field
- update ledger search payments with authorisation_summary field
- add ledger get one payment with authorisation_summary field


kbottla (approved):
LGTM!

PR #1245, opened 2021-09-21
PP-8695 Remove Jenkins deploy step

danworth:
CI Jenkins no longer needs to deploy Publicapi.

## WHAT YOU DID
Interestingly publicapi didn't seem to be checking pact compatibility before deployment.


oswaldquek (approved):
Wow did we not have a "check pact compatibility" here? 😮

PR #1281, opened 2021-11-24
PP-8844 - Remove synchronization for Redis calls

nimalank7:
Description:
- As described in the kick off the lettuce library used to talk to Redis already has multithreading support so no concurrency issues will arise
- We are removing the synchronization keyword to prevent publicapi processing requests one at a time in the event the Redis cluster goes offline

### DIFF
--- a/src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
+++ b/src/main/java/uk/gov/pay/api/filter/ratelimit/RedisRateLimiter.java
@@ -28,7 +28,7 @@ public class RedisRateLimiter {
     /**
      * @throws RateLimitException
      */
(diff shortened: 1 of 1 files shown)


danworth (approved):
This has been tested locally and in test-perf-1 to confirm that removing the `synchronised` of both `checkRateOf` and `updateAllowance` removes the bottleneck when Redis is unavailable. 

Further work is required to adjust the connection and command timeouts to further reduce the impact on clients when Redis is unavailable and Publicapi falls back to local rate limiting. With this change, Publicapi will continue to process requests concurrently but each request will incur an additional 2s latency whilst the Redis Command times out (or 10s latency if there is a connection timeout).

PR #1335, opened 2022-01-10
PP-8945 Allow connector pact tests to pass

stephencdaly:
The pact for getting a charge from connector simulates a scenario
where the payment is in "AWAITING CAPTURE REQUEST" status so that it
can check that a link to capture the payment exists.

It also checks for a fee and net amount, but we would not expect a
fee or net amount to exist on a charge in this state. As a result,
connector is returning a negative `net_amount` as we would expect for
failed payments which have a fee, whereas the pact test is expecting a
positive `net_amount` which will only be returned if the charge is in
a `CAPTURED` status.

Remove the check for `fee` and `net_amount` fields on this pact. Rename
the pact to indicate that it is for a charge in the capturable state.

Add a separate pact test that uses a connector state that has a captured
payment with fees to check the `fee` and `net_amount` fields.

Rename the pact for `with-delayed-capture-true` to
`publicapi-connector-get-created-charge` as the most important part of
this pact is it checks the next_url and next_url_post links are
returned. A separate pact test to just check the delayed capture field
is not required, as pacts shouldn't really test the exact values
returned, just the shape of the request.

### DIFF
--- a/src/test/java/uk/gov/pay/api/service/GetPaymentServiceTest.java
+++ b/src/test/java/uk/gov/pay/api/service/GetPaymentServiceTest.java
@@ -62,7 +62,7 @@ public class GetPaymentServiceTest {
     
     @Test
     @PactVerification({"connector"})
-    @Pacts(pacts = {"publicapi-connector-get-payment-with-delayed-capture-true"})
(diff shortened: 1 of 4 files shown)


SandorArpa (approved):
LOKTM

PR #1359, opened 2022-02-04
PP-9093 Remove explicit dependency on log4j

stephencdaly:
We don't bring in log4j transitively so can safely remove the explicit
dependency.


SandorArpa (approved):
LGTM

PR #1394, opened 2022-03-21
PP-9331: modify payments endpoint to support setting up payment instruments

jbz149:
- Added setup_agreement field to the Resources payload
- if setup_agreement, then forward agreement_id and save_paymentInstrument_to_agreement to Connector and PublicAI
- if setup_agreement, then forward agreement_id and save_paymentInstrument_to_agreement to Connector and Ledger
- added exception management and test cases

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/Charge.java
+++ b/src/main/java/uk/gov/pay/api/model/Charge.java
@@ -38,6 +38,8 @@ public class Charge {
     
     private boolean moto;
 
+    private String agreementId;
+
     private Long corporateCardSurcharge;
 
     private Long totalAmount;
@@ -60,7 +62,7 @@ public class Charge {
                   PaymentSettlementSummary settlementSummary, CardDetails cardDetails,
                   List<PaymentConnectorResponseLink> links, Long corporateCardSurcharge, Long totalAmount,
                   String gatewayTransactionId, ExternalMetadata metadata, Long fee, Long netAmount,
-                  AuthorisationSummary authorisationSummary) {
+                  AuthorisationSummary authorisationSummary, String agreementId) {
         this.chargeId = chargeId;
         this.amount = amount;
         this.state = state;
@@ -84,6 +86,7 @@ public class Charge {
         this.fee = fee;
         this.netAmount = netAmount;
         this.authorisationSummary = authorisationSummary;
+        this.agreementId = agreementId;
     }
 
     public static Charge from(ChargeFromResponse chargeFromResponse) {
@@ -110,7 +113,8 @@ public class Charge {
                 chargeFromResponse.getMetadata().orElse(null),
                 chargeFromResponse.getFee(),
                 chargeFromResponse.getNetAmount(),
-                chargeFromResponse.getAuthorisationSummary()
+                chargeFromResponse.getAuthorisationSummary(),
+                chargeFromResponse.getAgreementId()
         );
     }
 
@@ -138,8 +142,8 @@ public class Charge {
                 transactionResponse.getMetadata().orElse(null),
                 transactionResponse.getFee(),
                 transactionResponse.getNetAmount(),
-                transactionResponse.getAuthorisationSummary()
-        );
+                transactionResponse.getAuthorisationSummary(),
+                null);
     }
 
     public Optional<ExternalMetadata> getMetadata() {
@@ -236,5 +240,9 @@ public class Charge {
 
     public AuthorisationSummary getAuthorisationSummary() {
         return authorisationSummary;
+    }
+
+    public String getAgreementId() {
+        return agreementId;
     }
 }
--- a/src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
@@ -40,6 +40,7 @@ public class CreateCardPaymentRequest {
     public static final String PREFILLED_ADDRESS_COUNTRY_FIELD_NAME = "country";
     public static final String DELAYED_CAPTURE_FIELD_NAME = "delayed_capture";
     public static final String MOTO_FIELD_NAME = "moto";
+    public static final String SET_UP_AGREEMENT_FIELD_NAME = "set_up_agreement";
     public static final String SOURCE_FIELD_NAME = "source";
     public static final String METADATA = "metadata";
     public static final String INTERNAL = "internal";
@@ -81,6 +82,10 @@ public class CreateCardPaymentRequest {
     private final ExternalMetadata metadata;
     
     private final Internal internal;
+    
+    @JsonProperty("set_up_agreement")
+    @Size(min=26, max=26, message = "Field [set_up_agreement] length must be 26")
+    private String setUpAgreement;
 
     @Valid
     private final PrefilledCardholderDetails prefilledCardholderDetails;
@@ -97,6 +102,7 @@ public class CreateCardPaymentRequest {
         this.metadata = builder.getMetadata();
         this.prefilledCardholderDetails = builder.getPrefilledCardholderDetails();
         this.internal = builder.getInternal();
+        this.setUpAgreement = builder.getSetUpAgreement();
     }
     
     @Schema(description = "amount in pence", required = true, minimum = "1", maximum = "10000000", example = "12000")
@@ -165,6 +171,12 @@ public class CreateCardPaymentRequest {
         return Optional.ofNullable(internal);
     }
 
+    @JsonProperty("set_up_agreement")
+    @Schema(description = "agreement ID", required = false, example = "abcefghjklmnopqr1234567890", hidden = true)
+    public String getSetUpAgreement() {
+        return setUpAgreement;
+    }
+
     public String toConnectorPayload() {
         JsonStringBuilder request = new JsonStringBuilder()
                 .add("amount", this.getAmount())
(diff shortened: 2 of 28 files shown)


src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
```diff
@@ -189,6 +199,16 @@ public String toConnectorPayload() {
...
+        getSetUpAgreement().ifPresent(setUpAgreement -> {
+            request.add("agreement_id", this.getSetUpAgreement())
+                    .add("save_payment_instrument_to_agreement", true);
+        });
+
+        getSetUpAgreement().ifPresent(setUpAgreement -> {
+            request.add("agreement_id", this.getSetUpAgreement())
+                    .add("save_payment_instrument_to_agreement", true);
+        });
+
```
alexbishop1, line 211: Probably want to update the `toString()` method in this class too.

src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
```diff
@@ -111,6 +113,61 @@ public void createCardPaymentWithMetadata() {
...
+        publicAuthMockClient.mapBearerTokenToAccountId(API_KEY, GATEWAY_ACCOUNT_ID, CARD);
+
+        CreateChargeRequestParams createChargeRequestParams = aCreateChargeRequestParams()
+                .withAmount(100)
+                .withDescription(DESCRIPTION)
+                .withReference(REFERENCE)
+                .withReturnUrl(RETURN_URL)
+                .withAgreementId("12345678901234567890123456")
+                .withSavePaymentInstrumentToAgreement(true)
+                .withMetadata(Map.of("reconciled", true, "ledger_code", 123, "fuh", "fuh you"))
```
alexbishop1, line 127: We don’t really care about metadata stuff here, so kill it. What we want is something like `createCardPayment` with `.body("save_payment_instrument_to_agreement", is(true))` and `.body("agreement_id", is(whatever))`.

src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
```diff
@@ -584,7 +641,16 @@ public static String paymentPayload(CreateChargeRequestParams params) {
         params.getSource().ifPresent(source ->  {
             payload.addToNestedMap("source", source, "internal");
         });
-        
+
+        if (params.getAgreementId() != null) {
+            payload.add("agreement_id", params.getAgreementId());
+            payload.add("save_payment_instrument_to_agreement", true);
```
alexbishop1, line 647: I don’t think we want this line because the below `if` block will do it.

alexbishop1 (changes requested):
Looks good. Just a couple of things.

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
```diff
@@ -189,6 +199,16 @@ public String toConnectorPayload() {
...
         
+        getSetUpAgreement().ifPresent(setUpAgreement -> {
+            request.add("agreement_id", this.getSetUpAgreement())
+                    .add("save_payment_instrument_to_agreement", true);
+        });
+
+        getSetUpAgreement().ifPresent(setUpAgreement -> {
+            request.add("agreement_id", this.getSetUpAgreement())
+                    .add("save_payment_instrument_to_agreement", true);
+        });
```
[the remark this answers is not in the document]
jbz149: I don't thing we add it twice.
The name request is not from us.
it is just JsonStringBuilder which build from CreateCardPaymentRequest.

src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
```diff
@@ -111,6 +113,57 @@ public void createCardPaymentWithMetadata() {
...
+                .withReference(REFERENCE)
+                .withReturnUrl(RETURN_URL)
+                .withAgreementId("12345678901234567890123456")
+                .withSavePaymentInstrumentToAgreement(true)
+                .build();
+        connectorMockClient.respondOk_whenCreateCharge(GATEWAY_ACCOUNT_ID, createChargeRequestParams);
+
+        postPaymentResponse(paymentPayload(createChargeRequestParams))
+                .statusCode(HttpStatus.SC_CREATED)
+                .contentType(JSON);
```
alexbishop1, line 132: Probably want to add:

```java
.body("agreement_id", is("12345678901234567890123456"))
.body("save_payment_instrument_to_agreement", is(true))
```

src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
```diff
@@ -111,6 +113,57 @@ public void createCardPaymentWithMetadata() {
...
+        publicAuthMockClient.mapBearerTokenToAccountId(API_KEY, GATEWAY_ACCOUNT_ID, CARD);
+
+        CreateChargeRequestParams createChargeRequestParams = aCreateChargeRequestParams()
+                .withAmount(100)
+                .withDescription(DESCRIPTION)
+                .withReference(REFERENCE)
+                .withReturnUrl(RETURN_URL)
+                .withAgreementId("12345678901234567890123456")
+                .withSavePaymentInstrumentToAgreement(true)
+                .withMetadata(Map.of("reconciled", true, "ledger_code", 123, "fuh", "fuh you"))
```
alexbishop1, line 149: Kill the metadata from this test as well: it’s not relevant for what we’re testing.

alexbishop1 (changes requested):
Looks good. Just a couple of final things to do with the tests.

src/main/java/uk/gov/pay/api/model/Charge.java
```diff
@@ -86,6 +88,32 @@ public Charge(String chargeId, Long amount, PaymentState state, String returnUrl
         this.authorisationSummary = authorisationSummary;
     }
 
+    public Charge(String chargeId, Long amount, PaymentState state, String returnUrl, String description,
```
sfount, line 91: Any reason for a new constructor here rather than just adding agreement id to the previous constructor? It's a boxed type so there should be no issue in it being constructed with a `null` value if it hasn't been passed in

Having a new constructor for every new expected field doesn't seem scalable

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
```diff
@@ -165,6 +171,13 @@ public String getReturnUrl() {
         return Optional.ofNullable(internal);
     }
 
+
+    @JsonProperty("set_up_agreement")
+    @Schema(description = "agreement ID", required = false, example = "abcefghjklmnop1234567890", hidden = true)
```
sfount, line 176: Completely trivial but the example here is 24 characters and wouldn't actually pass the validation

sfount (approved):
Code and test coverage look reasonable to me, question around maintainable code on why a new constructor was added -- happy for that to discussed/ addressed here or in a follow up

sfount (approved):

alexbishop1 (changes requested):
Looks good. One small thing about an example and then there’s a change I don’t understand, which I’ve flagged up. Could you also squash the three commits into one?

src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
```diff
@@ -272,7 +372,7 @@ public void createCardPayment() {
                 .assertNotDefined("_links.events.type")
                 .assertNotDefined("_links.events.params");
 
-        connectorMockClient.verifyCreateChargeConnectorRequest(GATEWAY_ACCOUNT_ID, SUCCESS_PAYLOAD);
+        connectorMockClient.verifyCreateChargeConnectorRequest(GATEWAY_ACCOUNT_ID, CHARGE_REQUEST_PARAMS);
```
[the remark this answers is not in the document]
jbz149: We reverted it.

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
```diff
@@ -165,6 +171,12 @@ public String getReturnUrl() {
         return Optional.ofNullable(internal);
     }
 
+    @JsonProperty("set_up_agreement")
+    @Schema(description = "agreement ID", required = false, example = "abcefghjklmnopq1234567890", hidden = true)
```
alexbishop1, line 175: Still only 25 characters…

alexbishop1 (changes requested):
Needs more character.

(One more character to be exact. In the example agreement ID.)

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
jbz149: I think you are right, we forgot -n

echo -n abcefghjklmnopq1234567890 | wc -c
      25

src/main/java/uk/gov/pay/api/model/CreateCardPaymentRequest.java
```diff
@@ -165,6 +171,12 @@ public String getReturnUrl() {
         return Optional.ofNullable(internal);
     }
 
+    @JsonProperty("set_up_agreement")
+    @Schema(description = "agreement ID", required = false, example = "abcefghjklmnopqr1234567890", hidden = true)
```
alexbishop1, line 175: 1 … 2 … 3 … 4 … 5 … 6 … 7 … 8 … 9 … 10 … 11 … 12 … 13 … 14 … 15 … 16 … 17 … 18 … 19 … 20 … 21 … 22 … 23 … 24 … 25 … 26 — all good!

alexbishop1 (approved):

PR #1420, opened 2022-04-13
PP-9495 Accept description and user_identifier when creating agreements

sfount:
Accept and validation `description` and `user_identifier` fields on the
create agreement endpoint.

`description` is required and will throw an appropriate error if
missing.

`user_identifier` is optional but must be a string if set.

Note this does not extend the expected agreement repsonse from Connector
as Ledger will be used to serve agreements and this should be removed in
the next change.

### DIFF
--- a/src/main/java/uk/gov/pay/api/agreement/model/CreateAgreementRequest.java
+++ b/src/main/java/uk/gov/pay/api/agreement/model/CreateAgreementRequest.java
@@ -12,12 +12,22 @@ import java.util.Objects;
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class CreateAgreementRequest {
     
-    public static final int REFERENCE_MIN_LENGTH = 1;
-    public static final int REFERENCE_MAX_LENGTH = 255;
+    public static final String USER_IDENTIFIER_FIELD = "user_identifier";
+
+    public static final int MIN_LENGTH = 1;
+    public static final int MAX_LENGTH = 255;
 
     @JsonProperty("reference")
-    @Size(min= REFERENCE_MIN_LENGTH, max = REFERENCE_MAX_LENGTH, message = "Must be less than or equal to {max} characters length")
+    @Size(min= MIN_LENGTH, max = MAX_LENGTH, message = "Must be less than or equal to {max} characters length")
     private String reference;
+
+    @JsonProperty("description")
+    @Size(min= MIN_LENGTH, max = MAX_LENGTH, message = "Must be less than or equal to {max} characters length")
+    private String description;
+
+    @JsonProperty(USER_IDENTIFIER_FIELD)
+    @Size(min= MIN_LENGTH, max = MAX_LENGTH, message = "Must be less than or equal to {max} characters length")
+    private String userIdentifier;
 
     public CreateAgreementRequest() {
         // for Jackson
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/pay/api/it/CreateAgreementIT.java
```diff
@@ -66,12 +70,25 @@ public void shouldReturn400WhenWhenReferenceIsNull() {
                 .body("description", is("Missing mandatory attribute: reference"));
     }
 
+    @Test
+    public void shouldReturn400WhenWhenDescriptionIsNull() {
```
alexbishop1, line 74: Tests checking we reject an empty string and something that’s too long would be good.

alexbishop1, line 74: Tests for what happens if a `user_identifier` is null (not sure if we should reject it or ignore it), an empty string or too long would also make a fine addition.

alexbishop1 (changes requested):
Looks good but I think it could do with a few more tests.

src/test/java/uk/gov/pay/api/it/CreateAgreementIT.java
sfount: Yeah, I don't really like that these are done at an integration level or that they're done per property but I could increase the coverage here.

src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
```diff
@@ -117,7 +118,15 @@ static CreateCardPaymentRequest parsePaymentRequest(JsonNode paymentRequest) {
 
     static CreateAgreementRequest parseAgreementRequest(JsonNode agreementRequest) {
         var builder = CreateAgreementRequestBuilder.builder()
-                .reference(validateAndGetReference(agreementRequest));
+                .reference(validateAndGetReference(agreementRequest))
+                .description(validateAndGetDescription(agreementRequest));
+
+        if (agreementRequest.has(USER_IDENTIFIER_FIELD)) {
+            String userIdentifier = validateSkipNullValueAndGetString(agreementRequest.get(USER_IDENTIFIER_FIELD),
+                    aPaymentError(USER_IDENTIFIER_FIELD, CREATE_PAYMENT_VALIDATION_ERROR, "Field must be a string"));
```
alexbishop1, line 126: Not too sure about reusing `CREATE_PAYMENT_VALIDATION_ERROR` but we do such things elsewhere and I promised I’d create a story to review that, so I just did (PP-9498). So this can stay as it is for now.

src/test/java/uk/gov/pay/api/it/CreateAgreementIT.java
```diff
@@ -66,12 +70,87 @@ public void shouldReturn400WhenWhenReferenceIsNull() {
...
+        postAgreementRequest(agreementRequest.toConnectorPayload())
+                .statusCode(HttpStatus.SC_UNPROCESSABLE_ENTITY)
+                .contentType(JSON)
+                .body("field", is("description"))
+                .body("code", is("P0102"))
+                .body("description", is("Invalid attribute value: description. Must be less than or equal to 255 characters length"));
+    }
+
+    @Test
+    public void shouldReturn201WhenWhenOptionalUserIdentifierIsNull() {
```
alexbishop1, line 110: I assume this behaviour is consistent with other optional fields in our API.

src/main/java/uk/gov/pay/api/agreement/model/CreateAgreementRequest.java
```diff
@@ -12,41 +12,66 @@
...
-    @Size(min= REFERENCE_MIN_LENGTH, max = REFERENCE_MAX_LENGTH, message = "Must be less than or equal to {max} characters length")
+    @Size(min= MIN_LENGTH, max = MAX_LENGTH, message = "Must be less than or equal to {max} characters length")
     private String reference;
 
+    @JsonProperty("description")
+    @Size(min= MIN_LENGTH, max = MAX_LENGTH, message = "Must be less than or equal to {max} characters length")
+    private String description;
+
+    @JsonProperty(USER_IDENTIFIER_FIELD)
+    @Size(min= MIN_LENGTH, max = MAX_LENGTH, message = "Must be greater than or equal to {min} characters length and less than or equal to {max} characters length")
```
alexbishop1, line 29: It looks like similar fields use the message “Must be less than or equal to 255 characters length”, which we should probably match for consistency.

alexbishop1 (changes requested):
Just one thing about a validation message.

src/test/java/uk/gov/pay/api/it/CreateAgreementIT.java
sfount: Yes

src/main/java/uk/gov/pay/api/agreement/model/CreateAgreementRequest.java
alexbishop1: I agree it’s weird but I think consistency (even if it’s consistently weird!) is desirable here.

alexbishop1 (approved):

PR #1421, opened 2022-04-14
PP-9142 Accept authorisation_mode and return in responses.

stephencdaly:
Accept a new authorisation_mode property when creating a new payment.
Validate that this is a valid AuthorisationMode, and is one of the
allowed values - "web" or "moto_api". Ensure payments cannot be created
with `authorisation_mode` of "external", as this should only be set
internally for payments created using the telephone payment notification
API.

Return the `authorisation_mode` from connector/ledger in get and search
responses.

In the swagger file, hide the property on the create payment request as
we do not want people creating payments with this property set yet.
However, document the property on the payment response objects, as we
will start returning it immediately with a value of "web" for existing
payments.

### DIFF
--- a/openapi/publicapi_spec.json
+++ b/openapi/publicapi_spec.json
@@ -892,9 +892,9 @@
           },
           "card_type" : {
             "type" : "string",
-            "description" : "The card type, `credit` or `null` if not able to determine",
-            "enum" : [ "credit", "null" ],
-            "example" : "credit",
+            "description" : "The card type, `debit` or `credit` or `null` if not able to determine",
+            "enum" : [ "debit", "credit", "null" ],
+            "example" : "debit",
             "readOnly" : true
           },
           "cardholder_name" : {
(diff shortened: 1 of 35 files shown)


openapi/publicapi_spec.json
```diff
@@ -1110,11 +1110,19 @@
           "_links" : {
             "$ref" : "#/components/schemas/PaymentLinks"
           },
+          "agreement_id" : {
```
kbottla, line 1113: Are we fine to publish `agreement_id` ?

src/test/java/uk/gov/pay/api/json/CreateCardPaymentRequestDeserializerTest.java
```diff
@@ -16,6 +16,7 @@
 import uk.gov.pay.api.model.PrefilledCardholderDetails;
 import uk.gov.service.payments.commons.model.SupportedLanguage;
 
+import javax.ws.rs.WebApplicationException;
```
kbottla, line 19: Is this needed ?

stephencdaly: Looks like it isn't

openapi/publicapi_spec.json
stephencdaly: Oh I see this isn't our one

stephencdaly: Hidden

src/test/java/uk/gov/pay/api/json/CreateCardPaymentRequestDeserializerTest.java
stephencdaly: Removed

kbottla (approved):

PR #1428, opened 2022-04-25
PP-9390 pact test match type over value

nlsteers:
## WHAT YOU DID
Amend Connector-PublicAPI Pact test to match type over value


SandorArpa (approved):
LOKTM

PR #1436, opened 2022-04-28
PP-9407 implement request validation

SandorArpa:
## WHAT YOU DID

- implement validation of the request payload for AuthorisationRequest
- add relevant tests

with @stephencdaly

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/AuthorisationRequest.java
+++ b/src/main/java/uk/gov/pay/api/model/AuthorisationRequest.java
@@ -1,0 +1,49 @@
+package uk.gov.pay.api.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/api/model/AuthorisationRequest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+public class AuthorisationRequest {
+    @JsonProperty("one_time_token")
+    @NotNull
```
kbottla, line 11: Any reason `@NotNull` is used for all expect `cardholderName` which has `@NotBlank` annotation ?

SandorArpa: I've changed `oneTimeToken` to NotBlank also. The others has a minimum field size thus they only need null check.

kbottla (approved):

PR #1444, opened 2022-05-03
PP-9389 remove `return_url` from connector auth api pact tests

nlsteers:
## WHAT YOU DID
Remove the `return_url` check on auth API Pact tests as this is not required for this type of request


nlsteers:
Closed as duplicate

PR #1480, opened 2022-06-15
PP-9687 Refactor PaymentRefundsResource

stephencdaly:
Move code responsible for calling connector to create the refund and handling the response from connector into a service.

Co-authored-by: Alan Carter <alan.carter@digital.cabinet-office.gov.uk>


alan-gds (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay.

PR #1496, opened 2022-07-01
PP-9758: Add no cache directive

oswaldquek:
As per the IT health check (ITHC) report, add no cache directives for endpoints that
return sensitive data.

The ITHC report recommends to add `Cache-control: no-store` and `Pragma: no-cache` to `/v1/payments` and `/v1/payments/{paymentId}` (see JIRA ticket).


nlsteers (approved):
Looks good

PR #1506, opened 2022-07-15
PP-9812 Handle connector response for recurring payment request

alexbishop1:
When public API is handling a recurring payment request (`"authorisation_mode": "agreement"` and an `"agreement_id"`), we need to handle a couple of things differently depending on the response from connector.

If the payment is created successfully, we need to not include a cancel link in the public API response.

If connector responds with an error with the error identifier `AGREEMENT_NOT_FOUND`, public API needs to return a 400 response like this:

```json
{
  "field": "agreement_id",
  "code": "P0102",
  "description": "Invalid attribute value: agreement_id. Agreement does not exist"
}
```

If connector responds with an error with the error identifier `AGREEMENT_NOT_ACTIVE`, public API needs to return a 400 response like this:

```json
{
  "field": "agreement_id",
  "code": "P0102",
  "description": "Invalid attribute value: agreement_id. Agreement must be active"
}
```

with @sfount

### DIFF
--- a/src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
+++ b/src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
@@ -505,6 +505,52 @@ public class CreatePaymentIT extends PaymentResourceITestBase {
         connectorMockClient.verifyCreateChargeConnectorRequest(GATEWAY_ACCOUNT_ID, params);
     }
 
+    @Test
+    public void createPaymentWithAuthorisationModeAgreement() {
+        int amount = 1;
+
+        publicAuthMockClient.mapBearerTokenToAccountId(API_KEY, GATEWAY_ACCOUNT_ID);
+
+        connectorMockClient.respondOk_whenCreateCharge_withAuthorisationMode_Agreement(CHARGE_TOKEN_ID, GATEWAY_ACCOUNT_ID, aCreateOrGetChargeResponseFromConnector()
+                .withAmount(amount)
+                .withChargeId(CHARGE_ID)
+                .withState(CREATED)
+                .withDescription(DESCRIPTION)
+                .withReference(REFERENCE)
+                .withPaymentProvider(PAYMENT_PROVIDER)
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/pay/api/it/CreatePaymentIT.java
```diff
@@ -505,6 +505,52 @@ public void createPaymentWithAuthorisationModeMotoApi() {
...
+                .withRefundSummary(REFUND_SUMMARY)
+                .withCardDetails(CARD_DETAILS)
+                .withAuthorisationMode(AuthorisationMode.AGREEMENT)
+                .build());
+
+        CreateChargeRequestParams params = aCreateChargeRequestParams()
+                .withAmount(amount)
+                .withDescription(DESCRIPTION)
+                .withReference(REFERENCE)
+                .withAuthorisationMode(AuthorisationMode.MOTO_API)
```
sfount, line 534: I'm not quite sure why this is behaving as it is as we haven't supplied the appropriate authorisation mode here -- this should be changed in the next PR

src/test/java/uk/gov/pay/api/utils/mocks/ConnectorMockClient.java
```diff
@@ -319,6 +319,19 @@ public void respondOk_whenCreateCharge_withAuthorisationMode_MotoApi(String char
                         .withHeader(LOCATION, chargeLocation(gatewayAccountId, responseFromConnector.getChargeId()))
                         .withBody(buildChargeResponse(build)));
     }
+
+    public void respondOk_whenCreateCharge_withAuthorisationMode_Agreement(String chargeTokenId, String gatewayAccountId, ChargeResponseFromConnector responseFromConnector) {
+        ChargeResponseFromConnector build = aCreateOrGetChargeResponseFromConnector(responseFromConnector)
+                .withMoto(true)
```
sfount, line 325: This probably shouldn't be here

sfount (approved):

PR #1507, opened 2022-07-15
PP-9812 Send recurring payment request to connector

sfount:
Allows agreement authorisation mode.

Add all of the various additional builder/ parser/ validator params needed to pass a string through from the api to internal connector call.

Fix the test assertions so that it's validating the value would be propagated through to connector.

When creating a new payment, if `"authorisation_mode": "agreement"`, require an `"agreement_id"` (which must have a string value).

Conversely, if any other `"authorisation_mode"` is specified (or none at all), reject the request if it contains an `"agreement_id"`.

(The latter change will prevent surprising things happening if both `"set_up_agreement"` and `"agreement_id"` are specified.)

with @alexbishop1


alexbishop1 (commented):
Looks good but we should wait until PP-9880 is done on the connector side before merging this.

sfount (commented):
Code looks good to me 👍 some helpful refactors

sfount (commented):
Test refinement looks good 👍

alexbishop1 (approved):
We paired on parts of this and reviewed each other’s changes on the other parts.

PR #1508, opened 2022-07-18
PP-9812  "authorisation_mode": "agreement" requires "agreement_id"

alexbishop1:
When creating a new payment, if `"authorisation_mode": "agreement"`, require an `"agreement_id"` (which must have a string value).

Conversely, if any other `"authorisation_mode"` is specified (or none at all), reject the request if it contains an `"agreement_id"`.

(The latter change will prevent surprising things happening if both `"set_up_agreement"` and `"agreement_id"` are specified.)

with @sfount

Depends on #1507


alexbishop1:
Closing: we’re incorporating these changes into #1507.

PR #1510, opened 2022-07-19
PP-9784 Add cancel agreement endpoint

alexbishop1:
Add cancel agreement endpoint at `/v1/agreements/AGREEMENT_ID/cancel` that takes an empty POST body and, if successful, returns a 204.

with @sfount


sfount (commented):
This lines up with how the cancel charge functionality currently works

sfount (approved):
Looks good, we might want one happy path integration test to assert that appropriate thing on Connector actually gets called given a call to the API but given how simple this mapping is I think that can come with any additional required functionality

sfount:
Functionality depends on https://github.com/alphagov/pay-connector/pull/3901

sfount:
Review review: 
* The exception mapper should be registered with the public api app
* The cancel response should explicitly check for `204` from connector

PR #1513, opened 2022-07-20
PP-9545 Add get agreement from ledger

sfount:
* updates current agreement model to clarify that it is just the
  `ConnectorCreateAgreementResponse` serialiser that gets us the
  agreement ID
* adds Ledger methods to get an agreement
* fixtures and builders for future pacts and current IT


src/test/java/uk/gov/pay/api/it/CreateAgreementIT.java
```diff
@@ -39,8 +39,7 @@ public void shouldCreateAgreement() {
         postAgreementRequest(agreementPayload(createAgreementRequestParams))
                 .statusCode(HttpStatus.SC_CREATED)
                 .contentType(JSON)
-                .body("agreement_id", is(VALID_AGREEMENT_ID))
-                .body("reference", is(REFERENCE));
+                .body("agreement_id", is(VALID_AGREEMENT_ID));
```
alexbishop1, line 42: Why was reference removed?

alexbishop1 (changes requested):
Looks good. Couple of small things noted in the comments and do we need an `GetAgreementExceptionMapper`?

src/test/java/uk/gov/pay/api/it/CreateAgreementIT.java
sfount: Yeah good question. The idea will be that Connector returns just the resource ID of the thing it has created, the API can then look this up from the event store and respond.

That idea took a bit of time to develop and for some reason the agreement id and reference were parsed from Connector.

stephencdaly (dismissed):
Previous review comments have been addressed

PR #1516, opened 2022-07-22
PP-9675 implement validation of dispute search params

SandorArpa:
## WHAT YOU DID
- implement validating search params for disputes searches
- add test

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/RequestError.java
+++ b/src/main/java/uk/gov/pay/api/model/RequestError.java
@@ -86,7 +86,8 @@ public class RequestError {
         GET_AGREEMENT_LEDGER_ERROR("P0298", "Downstream system error"),
 
         CANCEL_AGREEMENT_CONNECTOR_BAD_REQUEST_ERROR("P0501", "Cancellation of agreement failed"),
-        CANCEL_AGREEMENT_CONNECTOR_ERROR("P0198", "Downstream system error");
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/api/model/RequestError.java
```diff
@@ -86,7 +86,8 @@ public enum Code {
         GET_AGREEMENT_LEDGER_ERROR("P0298", "Downstream system error"),
 
         CANCEL_AGREEMENT_CONNECTOR_BAD_REQUEST_ERROR("P0501", "Cancellation of agreement failed"),
-        CANCEL_AGREEMENT_CONNECTOR_ERROR("P0198", "Downstream system error");
+        CANCEL_AGREEMENT_CONNECTOR_ERROR("P0198", "Downstream system error"),
+        SEARCH_DISPUTES_VALIDATION_ERROR("P0401", "Invalid parameters: %s. See Public API documentation for the correct data formats");
```
stephencdaly, line 90: Would it make sense to put this next to the other error that uses code P0401 just so it's easier to see that this multiple use case exists.

Or as the message is the same, use the existing Enum value, possibly renaming it to make it more generic?

stephencdaly (approved):
Looks good, minor comment about the RequestError enum

src/main/java/uk/gov/pay/api/model/RequestError.java
SandorArpa: I'll move it in next iteration 👍

PR #1519, opened 2022-07-25
PP-9545 Make TransactionSearchResults generic

alexbishop1:
Transmogrify `TransactionSearchResults` into `SearchResults<E>` so we can use it for other kinds of search results.

with @sfount

### DIFF
--- a/src/main/java/uk/gov/pay/api/ledger/resource/TransactionsResource.java
+++ b/src/main/java/uk/gov/pay/api/ledger/resource/TransactionsResource.java
@@ -3,9 +3,10 @@ package uk.gov.pay.api.ledger.resource;
 import com.codahale.metrics.annotation.Timed;
 import io.dropwizard.auth.Auth;
 import uk.gov.pay.api.auth.Account;
+import uk.gov.pay.api.ledger.model.SearchResults;
 import uk.gov.pay.api.ledger.model.TransactionSearchParams;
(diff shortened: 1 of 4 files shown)


sfount (commented):
Should we add including the generics type for the warning case we found in the next commit? Might just be worth getting that in in case the next PR doesn't follow until tomorrow

Otherwise code looks good to me, happy to approve

src/main/java/uk/gov/pay/api/ledger/resource/TransactionsResource.java
```diff
@@ -30,9 +31,7 @@ public TransactionsResource(TransactionSearchService transactionSearchService) {
     @Timed
     @Path("/v1/transactions")
     @Produces(APPLICATION_JSON)
-    public TransactionSearchResults getTransactions(@Auth Account account,
-                                                    @BeanParam TransactionSearchParams searchParams) {
-        
+    public SearchResults<PaymentForSearchResult> getTransactions(@Auth Account account, @BeanParam TransactionSearchParams searchParams) {
```
sfount, line 34: Curiously this endpoint seems to return payments or refunds but squashes the transaction into a payment shape. This seems to largely work but doesn't make any sense. 

Making this a generic interface just points this out, the behaviour should continue

sfount (approved):

PR #1527, opened 2022-07-27
PP-9545 Add search agreements endpoint

alexbishop1:
Add `/v1/agreements` which maps to Ledger `/v1/api/agreement/`. Add search validator and basic coverage in line with other search endpoints.

with @sfount


src/test/java/uk/gov/pay/api/it/AgreementsIT.java
```diff
@@ -58,4 +58,45 @@ public void getAgreement_MissingAgreementShouldMapException() {
...
+                .withDescription("Description 2")
+                .withStatus("CANCELLED")
+                .withCreatedDate("2022-07-27T12:30:00Z")
+                .build();
+
+        ledgerMockClient.respondWithSearchAgreements(GATEWAY_ACCOUNT_ID, fixture1, fixture2);
+
+        given().port(app.getLocalPort())
+                .header(AUTHORIZATION, "Bearer " + PaymentResourceITestBase.API_KEY)
+                .get(AGREEMENTS_PATH)
```
sfount, line 81: Could we add something like page number and status to the query string just to assert that they get propagated all the way through from the call to the resource out to the wiremock ledger call? 

Even though it won't actually be filtering data I suppose it asserts that the optional params are correctly put into the hash map

src/main/java/uk/gov/pay/api/common/SearchConstants.java
```diff
@@ -6,6 +6,7 @@ public class SearchConstants {
     public static final String REFERENCE_KEY = "reference";
     public static final String EMAIL_KEY = "email";
     public static final String STATE_KEY = "state";
+    public static final String STATUS_KEY = "state";
```
sfount, line 9: I think this is the agreement property name

```suggestion
    public static final String STATUS_KEY = "status";
```

sfount (commented):
PR looks good, happy to approve. Question on if we could add one or two optional filter values to the wiremock conditions in the integration test to assert that they're correctly hooked up

src/test/java/uk/gov/pay/api/it/AgreementsIT.java
alexbishop1: Done this.

src/main/java/uk/gov/pay/api/common/SearchConstants.java
alexbishop1: Updating the test (as described in the previous comment) exposed this oversight and it’s been fixed.

src/main/java/uk/gov/pay/api/validation/AgreementSearchValidator.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private static void validateReference(String reference, List<String> validationErrors) {
+        if (isInvalid(reference, REFERENCE_MAX_LENGTH)) {
+            validationErrors.add("reference");
+        }
+    }
+
+    private static void validateStatus(String status, List<String> validationErrors) {
+        if (!(isBlank(status) || Arrays.stream(AgreementStatus.values()).anyMatch(validStatus -> validStatus.name().toLowerCase(ENGLISH).equals(status)))) {
```
sfount, line 63: As we'll be passing the status along as lowercase (ex. `created` vs. `CREATED`) we'll need to update Ledger to either match case insensitive or force lowercase values

sfount (approved):

PR #1528, opened 2022-07-27
PP-9662 Query payment refunds by transaction_type

kbottla:
## WHAT YOU DID
- Query payment refunds by transaction_type as the ledger endpoint /v1/transaction/{externalId}/transaction without query params will return refunds as well as disputes


SandorArpa (approved):
LOKTM

PR #1533, opened 2022-08-01
PP-9498 Update error codes for agreements API calls

alexbishop1:
Update the error codes for agreements API calls to match those we’ve agreed upon (which is essentially the same as the equivalent P0xxx error code for payments but P2xxx instead).


sfount (approved):
Looks great 🙆

PR #1539, opened 2022-08-09
PP-9498 Make cancel agreement handle error conditions

alexbishop1:
Correctly handle connector returning 404, 400 and 500 appropriately when cancelling an agreement and add tests to prove this.

### DIFF
--- a/src/main/java/uk/gov/pay/api/agreement/resource/AgreementsApiResource.java
+++ b/src/main/java/uk/gov/pay/api/agreement/resource/AgreementsApiResource.java
@@ -83,7 +83,7 @@ public class AgreementsApiResource {
 
     @POST
     @Path("/v1/agreements/{agreementId}/cancel")
-    @Consumes("application/json")
+    @Produces(APPLICATION_JSON)
     public Response createAgreement(@Parameter(hidden = true) @Auth Account account, @PathParam("agreementId") String agreementId) {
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/api/agreement/resource/AgreementsApiResource.java
```diff
@@ -83,7 +83,7 @@ public SearchResults<Agreement> getAgreements(@Auth Account account, @BeanParam
 
     @POST
     @Path("/v1/agreements/{agreementId}/cancel")
-    @Consumes("application/json")
+    @Produces(APPLICATION_JSON)
```
sfount, line 86: Is `204 NO CONTENT` strictly `application/json`? I'm not actually sure

sfount (commented):
Changes look good, very well covered 👍 . Happy to approve after a quick question about `@Produces` and the PR description

src/main/java/uk/gov/pay/api/agreement/resource/AgreementsApiResource.java
alexbishop1: This is what was causing me grief earlier. Normally there’s no content but when there’s an error there’s a JSON body so we need this or Jersey will panic and we end up with a 500.

sfount (approved):

PR #1575, opened 2022-09-29
Mirror the full format returned by the API in open api spec

acsauk:
## WHAT YOU DID
We're currently integrating with Gov UK Pay for a new service and we're using the open API spec as a seed for a mock service to run integration tests against using Stoplight Prism mock server. During development, I encountered a discrepancy regarding the date formats defined in the open API spec. The dates in the spec are in the format `2015-08-14T12:35:00Z` but when they are returned by the real API they're in the format `2022-09-29T12:20:03.873Z` (see sample JSON below)

```json
{
  "amount": 8200,
  "description": "A payment",
  "reference": "abc",
  "language": "en",
  "email": "a@b.com",
  "state": {
    "status": "created",
    "finished": false
  },
  "payment_id": "xxx",
  "payment_provider": "sandbox",
  "created_date": "2022-09-29T12:20:03.873Z",
  "refund_summary": {
    "status": "pending",
    "amount_available": 8200,
    "amount_submitted": 0
  },
  "settlement_summary": {},
  "delayed_capture": false,
  "moto": false,
  "return_url": "http://localhost:5050/payment-confirmation",
  "authorisation_mode": "web",
  "_links": {
    "self": {
      "href": "https://publicapi.payments.service.gov.uk/v1/payments/xxx",
      "method": "GET"
    },
    "next_url": {
      "href": "https://www.payments.service.gov.uk/secure/xxx",
      "method": "GET"
    },
    "next_url_post": {
      "type": "application/x-www-form-urlencoded",
      "params": {
        "chargeTokenId": "xxx"
      },
      "href": "https://www.payments.service.gov.uk/secure",
      "method": "POST"
    },
    "events": {
      "href": "https://publicapi.payments.service.gov.uk/v1/payments/xxx/events",
      "method": "GET"
    },
    "refunds": {
      "href": "https://publicapi.payments.service.gov.uk/v1/payments/xxx/refunds",
      "method": "GET"
    },
    "cancel": {
      "href": "https://publicapi.payments.service.gov.uk/v1/payments/xxx/cancel",
      "method": "POST"
    }
  }
}
```

Usually, this isn't a huge deal but Golang is pretty strict regarding the parsing dates and the date format we define needs to match exactly the format provided. Using the date format from the real API this matches up to [RFC3339Nano](https://pkg.go.dev/time#pkg-constants). It's possible to do some string manipulation to get it to match but ideally, we'd like to take the unedited value.

Given that the spec isn't 100% representative of real life I'm hoping this change won't be too controversial but happy to discuss it further.

### DIFF
--- a/openapi/publicapi_spec.json
+++ b/openapi/publicapi_spec.json
@@ -220,14 +220,14 @@
             "type" : "string"
           }
         }, {
-          "description" : "From date of payments to be searched (this date is inclusive). Example=2015-08-13T12:35:00Z",
+          "description" : "From date of payments to be searched (this date is inclusive). Example=2015-08-13T12:35:00.000Z",
           "in" : "query",
           "name" : "from_date",
           "schema" : {
             "type" : "string"
           }
         }, {
-          "description" : "To date of payments to be searched (this date is exclusive). Example=2015-08-14T12:35:00Z",
+          "description" : "To date of payments to be searched (this date is exclusive). Example=2015-08-14T12:35:00.000Z",
           "in" : "query",
           "name" : "to_date",
           "schema" : {
@@ -919,14 +919,14 @@
         "description" : "Search refunds by 'from' and 'to' date. The Authorisation token needs to be specified in the 'authorization' header as 'authorization: Bearer YOUR_API_KEY_HERE'",
         "operationId" : "Search refunds",
         "parameters" : [ {
-          "description" : "From date of refunds to be searched (this date is inclusive). Example=2015-08-13T12:35:00Z",
+          "description" : "From date of refunds to be searched (this date is inclusive). Example=2015-08-13T12:35:00.000Z",
           "in" : "query",
           "name" : "from_date",
           "schema" : {
             "type" : "string"
           }
         }, {
-          "description" : "To date of refunds to be searched (this date is exclusive). Example=2015-08-14T12:35:00Z",
+          "description" : "To date of refunds to be searched (this date is exclusive). Example=2015-08-14T12:35:00.000Z",
           "in" : "query",
           "name" : "to_date",
           "schema" : {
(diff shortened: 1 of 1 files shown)


openapi/publicapi_spec.json
```diff
@@ -1295,7 +1295,7 @@
           "created_date" : {
             "type" : "string",
             "description" : "The date and time the user's bank told GOV.UK Pay about this dispute.",
-            "example" : "2022-07-28T16:43:000Z",
+            "example" : "2022-07-28T16:43:00.000Z",
```
alexbishop1, line 1298: It looks like `2022-07-28T16:43:000Z` was clearly a mistake.

openapi/publicapi_spec.json
```diff
@@ -220,14 +220,14 @@
             "type" : "string"
           }
         }, {
-          "description" : "From date of payments to be searched (this date is inclusive). Example=2015-08-13T12:35:00Z",
+          "description" : "From date of payments to be searched (this date is inclusive). Example=2015-08-13T12:35:00.000Z",
```
alexbishop1, line 223: This is a bit more interesting: while we also include milliseconds for outputs in our API responses, this is an input we will allow something like `2015-08-13T12:35:00Z` there. Does that cause you problems?

alexbishop1 (commented):
Thanks for this. It looks like there’s definitely some errors in the existing file and we want to make it as easy as possible for you to integrate.

The spec is actually automatically generated from code annotations but we can take care of that.

There’s a question below that it would be helpful if you could answer.

alexbishop1:
Thanks again for reporting this. We’ve made some changes the example API output timestamps to match reality in https://github.com/alphagov/pay-publicapi/pull/1577. We have not changed the example API _input_ timestamps because as I said in an earlier comment we’re more flexible in the formats we accept there. Do let us know if you still have problems with the updated OpenAPI specification.

acsauk:
Hey Alex - so sorry for the nonresponse. I was away on annual leave just after I submitted the PR and then had a busy few weeks as soon as I got back. Appreciate you taking the changes into account. I'll go back and drop out the amended copy I made of the spec and pull it into prism directly from Github.

PR #1577, opened 2022-10-04
BAU correct swagger examples to the right format

SandorArpa:
## WHAT YOU DID
- there is a  discrepancy regarding the date formats defined in the open API spec. The dates in the spec are in the format 2015-08-14T12:35:00Z but when they are returned by the API they're in the format 2022-09-29T12:20:03.873Z

PR #1617, opened 2022-11-28
Nathan open api changes

NathanD-GDS:
OpenAPI changes for James

### DIFF
--- a/openapi/publicapi_spec.json
+++ b/openapi/publicapi_spec.json
@@ -1216,7 +1216,7 @@
           "created_date" : {
             "type" : "string",
             "description" : "The date you created the payment.",
-            "example" : "2016-01-21T17:15:00Z"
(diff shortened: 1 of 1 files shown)


openapi/publicapi_spec.json
```diff
@@ -1211,7 +1216,7 @@
           "created_date" : {
             "type" : "string",
             "description" : "The date you created the payment.",
-            "example" : "2016-01-21T17:15:00.000Z"
+            "example" : "2016-01-21T17:15:00Z"
```
kbottla, line 1219: Date formats shouldn't be changed. Existing examples in open API specs are in the correct format. See https://github.com/alphagov/pay-publicapi/pull/1577

NathanD-GDS: Yeah, not sure how this happened because I haven't touched the example values. I've pushed a change to fix this now after James pointed it out.

PR #1632, opened 2022-12-20
PP-10439 Handle 422 Recurring Payments Not Allowed response

james-peacock-gds:
Context: If connector receives a create agreement request from public_api for a gateway account that does not have recurring payments enabled, it will return a 422 response with the error_identifier RECURRING_CARD_PAYMENTS_NOT_ALLOWED

- Update CreateAgreementExceptionMapper to handle the 422 response from connector, sending a 422 response with the error identifier CREATE_AGREEMENT_RECURRING_CARD_PAYMENTS_NOT_ALLOWED
- Add unit tests

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/RequestError.java
+++ b/src/main/java/uk/gov/pay/api/model/RequestError.java
@@ -89,7 +89,7 @@ public class RequestError {
 
         CREATE_AGREEMENT_MISSING_FIELD_ERROR("P2101", "Missing mandatory attribute: %s"),
         CREATE_AGREEMENT_VALIDATION_ERROR("P2102", "Invalid attribute value: %s. %s"),
-        CREATE_AGREEMENT_RECURRING_CARD_PAYMENTS_NOT_ALLOWED_ERROR("P2103", "Recurring Card Payment agreements are not enabled for this account"),
+        CREATE_AGREEMENT_RECURRING_CARD_PAYMENTS_NOT_ALLOWED_ERROR("P2103", "Recurring card payments are currently disabled for this service. Contact support with your error code - https://www.payments.service.gov.uk/support/"),
         GET_AGREEMENT_NOT_FOUND_ERROR("P2200", "Not found"),
         GET_AGREEMENT_LEDGER_ERROR("P2298", "Downstream system error"),
 
--- a/src/main/java/uk/gov/pay/api/model/RequestError.java
+++ b/src/main/java/uk/gov/pay/api/model/RequestError.java
@@ -89,7 +89,7 @@ public class RequestError {
 
         CREATE_AGREEMENT_MISSING_FIELD_ERROR("P2101", "Missing mandatory attribute: %s"),
         CREATE_AGREEMENT_VALIDATION_ERROR("P2102", "Invalid attribute value: %s. %s"),
-
+        CREATE_AGREEMENT_RECURRING_CARD_PAYMENTS_NOT_ALLOWED_ERROR("P2103", "Recurring Card Payment agreements are not enabled for this account"),
         GET_AGREEMENT_NOT_FOUND_ERROR("P2200", "Not found"),
         GET_AGREEMENT_LEDGER_ERROR("P2298", "Downstream system error"),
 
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/api/service/AgreementServiceTest.java
```diff
@@ -82,6 +83,25 @@ void shouldCreateAgreement() {
...
+
+        CreateAgreementRequestParams createAgreementRequestParams = aCreateAgreementRequestParams()
+                .withReference(REFERENCE_ID)
+                .build();
+        when(mockInvocationBuilder.post(json(agreementPayload(createAgreementRequestParams)))).thenReturn(mockConnectorResponse);
+
+        when(mockConnectorResponse.getStatus()).thenReturn(HttpStatus.SC_UNPROCESSABLE_ENTITY);
+        CreateAgreementRequest agreementCreateRequest = new CreateAgreementRequest(CreateAgreementRequestBuilder
+                .builder().reference(REFERENCE_ID));
+        assertThrows(CreateAgreementException.class, () -> underTest.create(mockAccount, agreementCreateRequest));
```
stephencdaly, line 101: It looks like these tests will be pact tests eventually. Looking at the similar tests in `CreatePaymentServiceTest` we assert that the error identifier is passed through correctly from the connector response, and is the error identifier we expect in this scenario.

However, as this isn't currently a Pact test if we added that assertion we probably wouldn't be asserting much. So could possibly leave adding that assertion until we convert to pact tests..

src/main/java/uk/gov/pay/api/model/RequestError.java
```diff
@@ -89,7 +89,7 @@ public enum Code {
 
         CREATE_AGREEMENT_MISSING_FIELD_ERROR("P2101", "Missing mandatory attribute: %s"),
         CREATE_AGREEMENT_VALIDATION_ERROR("P2102", "Invalid attribute value: %s. %s"),
-
+        CREATE_AGREEMENT_RECURRING_CARD_PAYMENTS_NOT_ALLOWED_ERROR("P2103", "Recurring Card Payment agreements are not enabled for this account"),
```
NathanD-GDS, line 92: Is this the error message? If so, would it be possible to change it slightly?

> Recurring card payments are currently disabled for this service. Contact support with your error code - https://www.payments.service.gov.uk/support/

src/test/java/uk/gov/pay/api/exception/mapper/CreateAgreementExceptionMapperTest.java
```diff
@@ -36,4 +37,17 @@ void testExceptionMapping() {
                 is(expectedError.getDescription()));
         assertThat(returnedError.getCode(), is(expectedError.getCode()));
     }
+    
+    @Test
+    void testExceptionMappingForRecurringCardPaymentsNotAllowed() {
+        when(mockResponse.readEntity(ConnectorErrorResponse.class))
+                .thenReturn(new ConnectorErrorResponse(ErrorIdentifier.valueOf("RECURRING_CARD_PAYMENTS_NOT_ALLOWED"), null, null));
```
stephencdaly, line 44: Is there a good reason we lookup the enum values by a String value here and for the `RequestError.Code` rather than just using the enum constants directly? 

Looks like the pattern of doing this already exists for other tests, but I can't see why we'd do this. Maybe this was copied from a parameterised test at some point where it was necessary to look up from a string value?

stephencdaly (approved):

src/test/java/uk/gov/pay/api/service/AgreementServiceTest.java
james-peacock-gds: ok thanks for that

src/main/java/uk/gov/pay/api/model/RequestError.java
james-peacock-gds: yes will change, thanks

src/test/java/uk/gov/pay/api/exception/mapper/CreateAgreementExceptionMapperTest.java
james-peacock-gds: no idea - yes I followed what I saw elsewhere. Will adjust.

src/test/java/uk/gov/pay/api/service/AgreementServiceTest.java
james-peacock-gds: no changes made here, leave for pact tests as suggested

src/test/java/uk/gov/pay/api/exception/mapper/CreateAgreementExceptionMapperTest.java
james-peacock-gds: amended to refer to enum constants as suggested

stephencdaly (approved):
LGTM

PR #1642, opened 2023-01-09
BAU fix arm64 build failure

nlsteers:
## WHAT YOU DID

fixes an issue with the arm64 build process where some RUN commands were causing an error


katstevens (approved):
While the build works for me locally on master anyway, this branch a) also works b) is consistent with our AMD build.

PR #1658, opened 2023-02-03
PP-10488 Add missing logging for recurring payments

stephencdaly:
- Put the agreement_external_id on the MDC for requests having the agreementId as a path parameter
- Log a line when an agreement is created, including the agreement id


james-peacock-gds (approved):
LGTM

PR #1664, opened 2023-02-08
PP-10441 Return AgreementSearchResults when searching agreements

james-peacock-gds:
- previous updates to the openapi spec involved creating an AgreementSearchResults class purely
for the openapi spec
- update search agreements endpoint to return AgreementSearchResults instead of generic SearchResults
- correct naming and whitespace in AgreementSearchResults

### DIFF
--- a/openapi/publicapi_spec.json
+++ b/openapi/publicapi_spec.json
@@ -1378,6 +1378,7 @@
       },
       "Agreement" : {
         "type" : "object",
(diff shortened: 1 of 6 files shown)


openapi/publicapi_spec.json
```diff
@@ -1377,6 +1377,7 @@
       },
       "Agreement" : {
         "type" : "object",
+        "description" : "Contains agreements matching your search criteria.",
```
stephencdaly, line 1380: Hmm how has this one crept in 🤔

stephencdaly (approved):
LGTM!

PR #1706, opened 2023-04-20
PP-10713 Pact test for idempotency key create matching payment

SandorArpa:
## WHAT YOU DID

Add a Pact test to test the scenario where a second request is received to create a payment with an idempotency key and request body that matches that of an existing payment, which causes connector to return the original payment with a 200 response.

with @SandorArpa


james-peacock-gds (approved):
LGTM

PR #1714, opened 2023-04-27
PP-10844 Validate idempotency key

james-peacock-gds:
- Validate idempotency key to ensure only alphanumeric characters plus hyphens are used
- Respond with 422 and error code P0102
- Add integration test


SandorArpa (approved):
LGTM

PR #1721, opened 2023-05-12
BAU fix flaky tests

SandorArpa:
## WHAT YOU DID
The Idempotency-Key pattern matcher validataion is sometimes executed before the length check. This leads to a false error where the error `Header [Idempotency-Key] can only contain alphanumeric characters and hyphens` is returned.

- add allow empty string to regex


james-peacock-gds (approved):
LGTM

PR #1732, opened 2023-05-24
PP-10982 add new pact test

SandorArpa:
## WHAT YOU DID
- add a new pact test that verifies a rejected recurring payment with can retry true against connector


katstevens:
@SandorArpa can this PR be closed?

PR #1735, opened 2023-05-25
BAU Don't log service specified return url

sfount:
Don't include the return url when serialising card payments to be logged. 
These are API user specified and can contain GUIDs that look like card numbers.
The next URL can still be reviewed through internal tooling if needed on support.


SandorArpa (approved):
LOKTM

PR #1760, opened 2023-06-19
PP-10598-Pacts-for-cancel-agreement

james-peacock-gds:
- Add consumer test for cancel agreement
- Test for failed cancel agreement to be added in separate PR


SandorArpa (approved):
LOKTM

PR #1763, opened 2023-06-19
PP-10598 PACT test for cancelling a cancelled agreement

james-peacock-gds:
- Add consumer test publicapi-connector
- Attempt to cancel an already-cancelled agreement gets 400 response


SandorArpa (approved):
LGTM

PR #1771, opened 2023-06-22
PP-11188 refactor agreement service

SandorArpa:
## WHAT YOU DID
- combine all agreement related functions into AgreementService
- refactor AgreementsApiResource to use AgreementService only
- move relevant calls to ConnectorClient and LedgerClient
- update/rename relevant tests

with @james-peacock-gds


src/main/java/uk/gov/pay/api/agreement/resource/AgreementsApiResource.java
```diff
@@ -57,17 +53,20 @@ public class AgreementsApiResource {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AgreementsApiResource.class);
 
-    private final AgreementService agreementService;
-    private final LedgerService ledgerService;
+    private final CreateAgreementService createAgreementService;
```
kbottla, line 56: Any reason we need a different agreement service for each action? Can we not have one AgreementService with method for each action?

SandorArpa: We can, ofc. We were thinking that have everything like this, split into actions, eg `CreatePaymentService`, `CancelPaymentService`, `GetPaymentRefundService`, etc.

kbottla: Payments/refunds are a bit complex because of URI building and have additional logic to go to the connector or ledger so separate services classes are fine.

not sure if agreements need such granular services

src/main/java/uk/gov/pay/api/agreement/service/AgreementsService.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public Response cancelAgreement(Account account, String agreementId) {
+        Response connectorResponse = connectorService.cancelAgreement(account, agreementId);
+        if (connectorResponse.getStatus() != HttpStatus.SC_NO_CONTENT) {
+            throw new CancelAgreementException(connectorResponse);
+        }
+
+        connectorResponse.close();
+        return Response.noContent().build();
```
kbottla, line 51: Not for this PR. This method can return void instead of Response. I think it is ideal if the resource method does `Response.noContent().build();`.

src/main/java/uk/gov/pay/api/agreement/service/AgreementsService.java
```diff
@@ -0,0 +1,70 @@
...
+    public AgreementsService(ConnectorService connectorService, LedgerService ledgerService, PaginationDecorator paginationDecorator) {
+        this.connectorService = connectorService;
+        this.ledgerService = ledgerService;
+        this.paginationDecorator = paginationDecorator;
+    }
+
+    public AgreementCreatedResponse createAgreement(Account account, CreateAgreementRequest createAgreementRequest) {
+        var response = connectorService.createAgreement(account, createAgreementRequest);
+
+        if (response.getStatus() != HttpStatus.SC_CREATED) {
```
kbottla, line 38: This logic can potentially go into ConnectorService and the agreements service doesn't need to understand the `Response` object

src/main/java/uk/gov/pay/api/agreement/service/AgreementsService.java
```diff
@@ -0,0 +1,70 @@
...
+            throw new CreateAgreementException(response);
+        }
+        return response.readEntity(AgreementCreatedResponse.class);
+    }
+
+    public Response cancelAgreement(Account account, String agreementId) {
+        Response connectorResponse = connectorService.cancelAgreement(account, agreementId);
+        if (connectorResponse.getStatus() != HttpStatus.SC_NO_CONTENT) {
```
kbottla, line 46: Same as above

kbottla (approved):

src/main/java/uk/gov/pay/api/agreement/service/AgreementsService.java
SandorArpa: 👍

src/main/java/uk/gov/pay/api/agreement/service/AgreementsService.java
SandorArpa: 👍

src/main/java/uk/gov/pay/api/agreement/service/AgreementsService.java
SandorArpa: 👍

PR #1772, opened 2023-06-22
PP-11188 refactor agreement pact tests

SandorArpa:
## WHAT YOU DID
- remove AgreementService from the pact tests and use the actual service that interacts with the backend to run them
- refactor tests accordingly

with @james-peacock-gds


james-peacock-gds (approved):
LGTM
I'm confident that this change does not require further review and meets the secure coding requirements of [GOV.UK](http://gov.uk/) Pay

PR #1778, opened 2023-06-26
PP-10443 Add pact tests for create agreement endpoint

kbottla:
## WHAT YOU DID
- Added pact tests for the create agreements endpoint
- Fixed a few pact files which are causing issues when pacts are run as a provider locally.


SandorArpa (approved):
LOKTM

PR #1779, opened 2023-06-28
PP-10599 add a tekae a recurring payment pact test

SandorArpa:
## WHAT YOU DID
- add a new pact test for taking a recurring payment

with @james-peacock-gds


james-peacock-gds (approved):
I'm confident that this change does not require further review and meets the secure coding requirements of [GOV.UK](http://gov.uk/) Pay

PR #1780, opened 2023-06-29
PP-10599-PACT-for-setup-agreement

james-peacock-gds:
- Add PACT consumer test for setup agreement


SandorArpa (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #1781, opened 2023-06-29
Pp 11237

oswaldquek:
## WHAT YOU DID
Given CardPayment is under the uk.gov.pay.api.model package, having links is always going to be part of the API response for a card, and we are never going to use CardPayment as a database model, so we can move links information into it and delete a bunch of classes.
Payment can be deleted. We used to have a direct debit payment model, hence this super class but this is no longer needed.

Deleted the following classes:
PaymentWithAllLinks
Payment
PaymentLinksForSearch
GetPaymentResult - nothing special about this class, just extends CardPayment, so can be deleted.

### Documentation


heathd:
I don't know for _sure_ but I have a hunch that one of the reasons for the distinction between PaymentsWithAllLinks and without was to make it clear in auto-generated API docs that the _request_ payload didn't require to have links included.

I'm not sure now what Pay's API documentation strategy is but it would be worth double checking the consequences of this refactor on that? Did we move to having a separate curated api spec doc rather than using the auto-generated one?

openapi/publicapi_spec.json
```diff
@@ -2290,27 +2290,6 @@
...
-            "$ref" : "#/components/schemas/PostLink"
-          },
-          "events" : {
-            "$ref" : "#/components/schemas/Link"
-          },
-          "refunds" : {
-            "$ref" : "#/components/schemas/Link"
-          },
-          "self" : {
-            "$ref" : "#/components/schemas/Link"
```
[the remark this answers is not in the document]
oswaldquek: Having run a test it looks like more links will be included, but they will be null where applicable:
```
            "_links": {
                "self": {
                    "href": "http://publicapi.url/v1/payments/0",
                    "method": "GET"
                },
                "next_url": null,
                "next_url_post": null,
                "auth_url_post": null,
                "events": {
                    "href": "http://publicapi.url/v1/payments/0/events",
                    "method": "GET"
                },
                "refunds": {
                    "href": "http://publicapi.url/v1/payments/0/refunds",
                    "method": "GET"
                },
                "cancel": {
                    "href": "http://publicapi.url/v1/payments/0/cancel",
                    "method": "POST"
                },
                "capture": null
```
A few things to note here:
* [HATEOAS](https://en.wikipedia.org/wiki/HATEOAS) states that: "With HATEOAS, a client interacts with a network application whose application servers provide information dynamically through hypermedia. A REST client needs little to no prior knowledge about how to interact with an application or server beyond a generic understanding of hypermedia." It follows that having a nulls for `next_url`, `next_url_post`, and `auth_url_post` are ok because the client will explicitly know that those links don't exist.
* It seems like we're already returning a [capture link, which could be null](https://github.com/alphagov/pay-publicapi/blob/master/src/main/java/uk/gov/pay/api/model/links/PaymentLinksForSearch.java#L31), so we're already returning values for links.
* I could exclude null fields by using `@JsonInclude(Include.NON_NULL)` (haven't tried at time of writing as it's near the end of the day.)

oswaldquek:
> I don't know for _sure_ but I have a hunch that one of the reasons for the distinction between PaymentsWithAllLinks and without was to make it clear in auto-generated API docs that the _request_ payload didn't require to have links included.
> 
> I'm not sure now what Pay's API documentation strategy is but it would be worth double checking the consequences of this refactor on that? Did we move to having a separate curated api spec doc rather than using the auto-generated one?

PaymentsWithAllLinks are only used in responses and doesn't affect requests. I believe we're using an [auto-generated spec](https://github.com/alphagov/pay-publicapi/blob/master/pom.xml#L503-L513).

openapi/publicapi_spec.json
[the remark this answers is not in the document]
oswaldquek: Am not fussed though, happy to keep PaymentLinksForSearch if you feel it's better.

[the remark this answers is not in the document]
oswaldquek: `@JsonInclude(JsonInclude.Include.NON_EMPTY)` on `PaymentLinks` class will ensure empty values don't get returned.

src/test/java/uk/gov/pay/api/resources/PaymentsResourceCreatePaymentTest.java
```diff
@@ -95,14 +96,14 @@ void createNewPayment_withCardPayment_invokesCreatePaymentService() {
...
+        assertThat(entity, sameInstance(injectedResponse));
     }
 
     @NotNull
-    private PaymentWithAllLinks aSuccessfullyCreatedPayment() {
-        final Address cardholderAddress = new Address("123 Acacia Ave", "", "", "London", "GB");
-        return new PaymentWithAllLinks(
-                "abc123",
+    private CardPayment aSuccessfullyCreatedPayment() {
+        //TODO use builder pattern
```
oswaldquek, line 105: Using the builder pattern can be in another PR.

src/main/java/uk/gov/pay/api/model/links/PaymentLinks.java
```diff
@@ -12,6 +12,7 @@
 import static javax.ws.rs.HttpMethod.POST;
 
 @Schema(name = "PaymentLinks", description = "links for payment")
+@JsonInclude(value = JsonInclude.Include.NON_EMPTY)
```
stephencdaly, line 15: Might we have to be cautious with this, as it changes the API behaviour.

Currently we do include the capture and cancel links in search results if they are null
```
            "_links": {
                "self": {
                    "href": "https://publicapi.pymnt.uk/v1/payments/1jvgk16f8j1q8t1ec8pv9mnbbj",
                    "method": "GET"
                },
                "cancel": null,
                "events": {
                    "href": "https://publicapi.pymnt.uk/v1/payments/1jvgk16f8j1q8t1ec8pv9mnbbj/events",
                    "method": "GET"
                },
                "refunds": {
                    "href": "https://publicapi.pymnt.uk/v1/payments/1jvgk16f8j1q8t1ec8pv9mnbbj/refunds",
                    "method": "GET"
                },
                "capture": null
            },
```

But we don't include for get single payment:

```
    "_links": {
        "self": {
            "href": "https://publicapi.pymnt.uk/v1/payments/b930o02iq7nfdvfnk7opnkofsj",
            "method": "GET"
        },
        "events": {
            "href": "https://publicapi.pymnt.uk/v1/payments/b930o02iq7nfdvfnk7opnkofsj/events",
            "method": "GET"
        },
        "refunds": {
            "href": "https://publicapi.pymnt.uk/v1/payments/b930o02iq7nfdvfnk7opnkofsj/refunds",
            "method": "GET"
        }
    }
```

I'm not quite sure the mechanism for excluding the null links for GET single payment responses if this annotation wasn't present before, but is it safest to ensure the behaviour is unchanged?

oswaldquek: > Currently we do include the capture and cancel links in search results if they are null

I feel like this wouldn't be a problem. If they can be null, any client would have to test that they are there to begin with.

oswaldquek: Despite all the above, I'm happy to revert the deletion of `PaymentLinksForSearch` if you think that's safer!

stephencdaly: I think we've had concerns in the past about making similar changes. I think there is a possibility for consumers to be doing interesting things that might break if we stop including properties when they're null when we previously including them. It possibly constitutes a breaking change to the API.
If reverting the deletion of `PaymentLinksForSearch` still allows you to improve things in some way, maybe that would be easiest.

oswaldquek: Happy to do that, will revert.

oswaldquek: Ok, [done](https://github.com/alphagov/pay-publicapi/pull/1781/commits/9274f5775bc146d1ba73380aed0e0d2702c9afda).

PR #1787, opened 2023-07-05
PP-11237: Remove redundant classes

oswaldquek:
Payment can be deleted. We used to have a direct debit payment model, hence this super class but this is no longer needed.

GetPaymentResult - nothing special about this class, just extends CardPayment, so can be deleted and replaced with PaymentWithAllLinks.


openapi/publicapi_spec.json
```diff
@@ -1552,6 +1552,135 @@
           }
         }
       },
+      "CardPayment" : {
```
stephencdaly, line 1555: GetPaymentResult included a "_links" property. Looks like this was only used fro the API spec. Can we use PaymentWithAllLinks for the API spec or is there some reason this doesn't work?

stephencdaly (approved):
LGTM!

PR #1793, opened 2023-07-06
bau: Test and PaymentWithAllLinks constructor cleanup

oswaldquek:
Removed JsonAssert usages in PaymentsResourceSearchIT as they weren't doing anything. Try adding the following in
`PaymentsResourceSearchIT.searchPayments_shouldOnlyReturnAllowedProperties` in the current master branch:
```
JsonAssert.with(responseBody).assertNotDefined("_links.refunds.href")
```

It should cause the test to fail but it doesn't.

Also use a builder pattern for PaymentWithAllLinks.

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
+++ b/src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
@@ -25,7 +25,7 @@ public class PaymentWithAllLinks extends CardPayment {
         return links;
     }
 
-    public PaymentWithAllLinks(String chargeId, long amount, PaymentState state, String returnUrl, String description,
+    private PaymentWithAllLinks(String chargeId, long amount, PaymentState state, String returnUrl, String description,
                                String reference, String email, String paymentProvider, String createdDate, SupportedLanguage language,
                                boolean delayedCapture, boolean moto, RefundSummary refundSummary, PaymentSettlementSummary settlementSummary, CardDetails cardDetails,
                                List<PaymentConnectorResponseLink> paymentConnectorResponseLinks, URI selfLink, URI paymentEventsUri, URI paymentCancelUri,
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
```diff
@@ -100,4 +101,201 @@ public static PaymentWithAllLinks getPaymentWithLinks(
...
+        private Long totalAmount;
+        private String providerId;
+        private ExternalMetadata metadata;
+        private Long fee;
+        private Long netAmount;
+        private AuthorisationSummary authorisationSummary;
+        private String agreementId;
+        private AuthorisationMode authorisationMode;
+
+        public PaymentWithAllLinksBuilder setChargeId(String chargeId) {
```
stephencdaly, line 138: the builder pattern usually has method prefix `with...` in our repos

src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
```diff
@@ -100,4 +101,201 @@ public static PaymentWithAllLinks getPaymentWithLinks(
...
+            this.agreementId = agreementId;
+            return this;
+        }
+
+        public PaymentWithAllLinksBuilder setAuthorisationMode(AuthorisationMode authorisationMode) {
+            this.authorisationMode = authorisationMode;
+            return this;
+        }
+
+        public PaymentWithAllLinks createPaymentWithAllLinks() {
```
stephencdaly, line 293: should this method be named `build` to be consistent with other builders?

stephencdaly (approved):
Looks good, just minor style comments

src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
oswaldquek: ah...does it really matter? This was auto-generated by Intellij...

src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
oswaldquek: yeah I'll make it consistent. i had let intellij do all the auto-generation here.

src/main/java/uk/gov/pay/api/model/links/PaymentWithAllLinks.java
stephencdaly: Usually it lets you select which one to use doesn't it? or maybe I have a specific plugin installed to generate builders

Consistency is nice just because if you expect one thing because its like that everywhere else but its different, it's just a slight annoyance in your day.

stephencdaly: Ah I have this plugin installed https://plugins.jetbrains.com/plugin/6585-builder-generator

stephencdaly (approved):

PR #1800, opened 2023-07-21
PP-11223: Set up prometheus metrics

oswaldquek:
Set up a prometheus metric endpoint /metrics.
Append ECS metadata info (via the PrometheusDefaultLabelSampleBuilder class in pay-java-commons) as default labels.

Analogus to alphagov/pay-connector#4625, which has been tested.


nlsteers (approved):
🥇

PR #1802, opened 2023-07-24
PP-11223: Fix deploy

oswaldquek:
Publicapi failed to deploy due to:
```
at uk.gov.pay.api.app.PublicApi.main(PublicApi.java:220)
Caused by: java.lang.ClassNotFoundException: org.json.JSONException
```
The transitive org.json:json dependency from pay-java-commons was not included due to the test scope overriding it in publicapi's pom.xml


nlsteers (approved):
🚑

PR #1833, opened 2023-09-22
PP-11511 surface `wallet_type` in charge responses

nlsteers:
## WHAT YOU DID

- renamed the `CardDetails` class to `CardDetailsFromResponse` to better distinguish between deserialising and serialising DTOs
- added new class `CardDetails` that surfaces `wallet_type` information when serialising reponses
- added new enum `Wallet`
- added new util for deserialising `wallet_type` from Ledger/Connector responses to Wallet enum
- added integration and pact tests for new scenarios


src/main/java/uk/gov/pay/api/model/CardDetails.java
```diff
@@ -36,6 +38,9 @@ public class CardDetails {
     @JsonProperty("card_type")
     private final String cardType;
 
+    @JsonProperty("wallet_type")
+    private String walletType;
```
stephencdaly, line 42: The only thing I'm slightly unsure about is that we use this class for both deserialisation of the response from connector/ledger and serialisation of the public API response. I wonder whether this is slightly confusing as someone might expect this field to be on the connector/ledger response.

I wonder whether it's clearer if we create a separate class to represent the public API response, and have a constructor method that accepts a CardDetails and the wallet type.

openapi/publicapi_spec.json
```diff
@@ -1549,6 +1549,12 @@
             "type" : "string",
             "example" : "1234",
             "readOnly" : true
+          },
+          "wallet_type" : {
+            "type" : "string",
+            "description" : "The digital wallet type that the user paid with",
+            "enum" : [ "APPLE_PAY", "GOOGLE_PAY" ],
```
stephencdaly, line 1556: Not sure how this `enum` field appears

src/main/java/uk/gov/pay/api/model/CardDetails.java
nlsteers: I think it's confusing too, I've added a new class `CardDetailsFromResponse` to better distinguish between which DTOs are used for deserialisation/serialisation.

openapi/publicapi_spec.json
nlsteers: have updated the schema deff in the `CardDetails` class with allowable values, still shows up as an enum though 🤷🏻‍♂️

stephencdaly: It's correct now, which is what matters. Before it wasn't correct, but I couldn't see how it was being generated at all

stephencdaly (dismissed):
Looks good!

stephencdaly (approved):

PR #1937, opened 2024-02-21
ZD-5631868 Add payment_id to refund OpenAPI spec

alexbishop1:
The `payment_id` field is included in the refund object and documented at https://docs.payments.service.gov.uk/api_reference/search_refunds_reference/#attributes-you-ll-get-in-a-search-refunds-response but it is not included in our OpenAPI spec.

Since it’s already there and documented, people will be relying on it so the only sensible thing to do is to include it in the OpenAPI spec.

Description nabbed from the documentation and is identical to that used elsewhere in the OpenAPI spec.

### DIFF
--- a/openapi/publicapi_spec.json
+++ b/openapi/publicapi_spec.json
@@ -2500,6 +2500,12 @@
             "example" : "2017-01-10T16:52:07.855Z",
             "readOnly" : true
           },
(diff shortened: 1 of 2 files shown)


nlsteers (approved):
🥇

PR #1956, opened 2024-03-21
bau: Make publicauth endpoint clear

oswaldquek:
This will save some effort looking into terraform to see what the value for the PUBLIC_AUTH_URL env var is.


james-peacock-gds (approved):
LGTM

PR #2010, opened 2024-06-11
PP-12517: Distinguish between any exceptions caused by increment and …

oswaldquek:
…expire

This might help us understand a bit better when the RedisException occurs.

The changing of the logged messages isn't going to break any splunk alerts and that's alerted on a [different string](https://github.com/alphagov/cyber-security-splunk-apps/blob/master/gds-004-pay/default/savedsearches.conf#L1334).


james-peacock-gds (approved):
LGTM

PR #2078, opened 2024-09-24
PP-13201 implement AuthorisationSummary.ThreeDSecure not required

SandorArpa:
## WHAT YOU DID

Only include authorisation summary in public API responses if 3D Secure was required.

- add a helper class that determines whether we need to include AuthorisationSummary to the return API call objects

### DIFF
--- a/src/main/java/uk/gov/pay/api/model/search/card/PaymentForSearchResult.java
+++ b/src/main/java/uk/gov/pay/api/model/search/card/PaymentForSearchResult.java
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.v3.oas.annotations.media.Schema;
 import uk.gov.pay.api.model.AuthorisationSummary;
 import uk.gov.pay.api.model.CardDetails;
-import uk.gov.pay.api.model.CardDetailsFromResponse;
 import uk.gov.pay.api.model.CardPayment;
 import uk.gov.pay.api.model.PaymentConnectorResponseLink;
 import uk.gov.pay.api.model.PaymentSettlementSummary;
@@ -12,6 +11,7 @@ import uk.gov.pay.api.model.PaymentState;
 import uk.gov.pay.api.model.RefundSummary;
 import uk.gov.pay.api.model.TransactionResponse;
 import uk.gov.pay.api.model.links.PaymentLinksForSearch;
+import uk.gov.pay.api.utils.AuthorisationSummaryHelper;
 import uk.gov.service.payments.commons.model.AuthorisationMode;
 import uk.gov.service.payments.commons.model.SupportedLanguage;
 import uk.gov.service.payments.commons.model.charge.ExternalMetadata;
@@ -85,7 +85,7 @@ public class PaymentForSearchResult extends CardPayment {
                 paymentResult.getMetadata().orElse(null),
                 paymentResult.getFee(),
                 paymentResult.getNetAmount(),
-                paymentResult.getAuthorisationSummary(),
+                AuthorisationSummaryHelper.includeAuthorisationSummaryWhen3dsRequired(paymentResult.getAuthorisationSummary()),
                 paymentResult.getAuthorisationMode());
     }
 
--- a/src/main/java/uk/gov/pay/api/utils/AuthorisationSummaryHelper.java
+++ b/src/main/java/uk/gov/pay/api/utils/AuthorisationSummaryHelper.java
@@ -1,0 +1,14 @@
+package uk.gov.pay.api.utils;
+
+import uk.gov.pay.api.model.AuthorisationSummary;
+
+public class AuthorisationSummaryHelper {
+    public static AuthorisationSummary includeAuthorisationSummaryWhen3dsRequired(AuthorisationSummary authorisationSummary) {
+        if (authorisationSummary != null &&
+                authorisationSummary.getThreeDSecure() != null &&
+                authorisationSummary.getThreeDSecure().isRequired()) {
+            return authorisationSummary;
+        }
+        return null;
+    }
+}
--- a/src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
+++ b/src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
@@ -694,7 +694,7 @@ public class PaymentsResourceGetIT extends PaymentResourceITestBase {
     }
 
     @Test
-    public void getPaymentWithAuthorisationSummaryThroughLedger() {
+    public void getPaymentWithAuthorisationSummaryAndThreeDSecureRequiredIsTrueThroughLedger() {
         AuthorisationSummary authorisationSummary = new AuthorisationSummary(new ThreeDSecure(true));
         ledgerMockClient.respondWithTransaction(CHARGE_ID,
                 getLedgerTransaction()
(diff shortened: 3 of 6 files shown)


src/main/java/uk/gov/pay/api/model/search/card/PaymentForSearchResult.java
```diff
@@ -85,7 +85,10 @@ public static PaymentForSearchResult valueOf(
                 paymentResult.getMetadata().orElse(null),
                 paymentResult.getFee(),
                 paymentResult.getNetAmount(),
-                paymentResult.getAuthorisationSummary(),
+                paymentResult.getAuthorisationSummary() == null ?
+                        null : paymentResult.getAuthorisationSummary().getThreeDSecure() == null ?
+                        null : paymentResult.getAuthorisationSummary().getThreeDSecure().isRequired() ?
+                        paymentResult.getAuthorisationSummary() : null,
```
alexbishop1, line 91: This is the third time this logic is repeated — could it be encapsulated in a helper method? I think even a static method somewhere would be preferable to this repetition.

src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
```diff
@@ -707,6 +707,20 @@ public void getPaymentWithAuthorisationSummaryThroughLedger() {
         assertPaymentWithAuthorisationSummary(response);
     }
 
+    @Test
+    public void getPaymentWithAuthorisationSummaryAndThreeDSecureFalseThroughLedger() {
```
alexbishop1, line 711: This name is inconsistent with the others (in terms of whether or not there’s `No` and `Required`).

src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
```diff
@@ -765,6 +778,21 @@ public void getPaymentWithNoAuthorisationSummaryThroughConnector() {
         response.body("authorisation_summary", is(nullValue()));
     }
 
+    @Test
+    public void getPaymentWithNoAuthorisationSummaryThroughConnectorWhenRequiredThreeDSecureIsFalse() {
```
alexbishop1, line 782: This name is inconsistent with the others (in terms of whether or not there’s `No` and `Required`).

src/test/java/uk/gov/pay/api/it/ledger/TransactionsResourceIT.java
```diff
@@ -139,6 +139,38 @@ public void shouldReturnAListOfTransactions() {
                 .body("_links.last_page.href", is(expectedChargesLocationFor("?reference=reference&display_size=500&page=1")));
     }
 
+    @Test
+    public void shouldReturnAListOfTransactionsWithoutAuthorisationSummary() {
```
alexbishop1, line 143: Maybe `shouldReturnAListOfTransactionsWithoutAuthorisationSummaryWhenThreeDSecureRequiredFalse`?

alexbishop1 (changes requested):
Not pretty but it’ll work. I think we might be able to encapsulate some of the repeated logic. Some of the test names are inconsistent.

src/main/java/uk/gov/pay/api/utils/AuthorisationSummaryHelper.java
```diff
@@ -0,0 +1,12 @@
...
+
+import uk.gov.pay.api.model.AuthorisationSummary;
+
+public class AuthorisationSummaryHelper {
+    public static AuthorisationSummary includeAuthorisationSummaryWhen3dsRequired(AuthorisationSummary authorisationSummary) {
+        if (authorisationSummary == null || authorisationSummary.getThreeDSecure() == null || !authorisationSummary.getThreeDSecure().isRequired()) {
+            return null;
+        }
+        return authorisationSummary;
+    }
```
alexbishop1, line 11: This might be a bit clearer if it’s inverted:

```java
if (authorisationSummary != null && authorisationSummary.getThreeDSecure() != null && authorisationSummary.getThreeDSecure().isRequired()) {
    return authorisationSummary;
}
return null;
```

But I think the difference is fairly marginal so I’m not going to insist on it.

src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
```diff
@@ -765,6 +778,21 @@ public void getPaymentWithNoAuthorisationSummaryThroughConnector() {
         response.body("authorisation_summary", is(nullValue()));
     }
 
+    @Test
+    public void getPaymentShouldNotIncludeNoAuthorisationSummaryThroughConnectorWhenThreeDSecureRequiredIsFalse() {
```
alexbishop1, line 782: The double negative `ShouldNotIncludeNo` doesn’t look right.

src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
```diff
@@ -765,6 +778,21 @@ public void getPaymentWithNoAuthorisationSummaryThroughConnector() {
...
+                        .withAuthorisationSummary(authorisationSummary)
+                        .build());
+
+        ValidatableResponse response = getPaymentResponse(CHARGE_ID);
+
+        assertCommonPaymentFields(response);
+        assertConnectorOnlyPaymentFields(response);
+        response.body("authorisation_summary", is(nullValue()));
+    }
+
```
alexbishop1, line 795: Am I right in thinking existing tests cover there being no `AuthorisationSummary` at all (essentially that the null checks are in place)?

src/test/java/uk/gov/pay/api/utils/AuthorisationSummaryHelperTest.java
```diff
@@ -0,0 +1,35 @@
...
+        AuthorisationSummary authorisationSummary = new AuthorisationSummary(new ThreeDSecure(false));
+        assertThat(includeAuthorisationSummaryWhen3dsRequired(authorisationSummary), is(nullValue()));
+    }
+
+    @Test
+    void shouldReturnAuthorisationSummaryWhenThreeDSecureRequiredIsTrue() {
+        AuthorisationSummary authorisationSummary = new AuthorisationSummary(new ThreeDSecure(true));
+        assertThat(includeAuthorisationSummaryWhen3dsRequired(authorisationSummary), is(authorisationSummary));
+    }
+}
```
alexbishop1, line 35: Comprehensive.

alexbishop1 (changes requested):
Not normally a fan of static methods in utility classes but I think it fits the bill in this case. One test name that does not look quite right but otherwise we’re about good to go.

src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
SandorArpa: Good question, I'll check

src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
SandorArpa: But I like double negatives :(

src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
SandorArpa: Yes, I added it in 2021 `getPaymentWithNoAuthorisationSummaryThroughConnector`

SandorArpa: And now added `getPaymentWithNoAuthorisationSummaryThroughLedger`

src/main/java/uk/gov/pay/api/utils/AuthorisationSummaryHelper.java
SandorArpa: Inverted

alexbishop1 (approved):

PR #2116, opened 2024-11-26
PP-13315: Deal with NON_HTTPS_RETURN_URL_NOT_ALLOWED_FOR_A_LIVE_ACCOUNT

oswaldquek:
Deal with possible NON_HTTPS_RETURN_URL_NOT_ALLOWED_FOR_A_LIVE_ACCOUNT when creating a charge.

This commit also upgrades CreatePaymentWithHttpReturnUrlIT to a junit5 test. As part of this MockHelperFunctions is created so WireMockExtension (junit5) and WireMockClassRule (junit4) can be DRY.

### DIFF
--- a/src/test/java/uk/gov/pay/api/it/rule/RedisDockerRule.java
+++ b/src/test/java/uk/gov/pay/api/it/rule/RedisDockerRule.java
@@ -1,5 +1,7 @@
 package uk.gov.pay.api.it.rule;
 
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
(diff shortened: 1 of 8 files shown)


oswaldquek:
This PR enables https://github.com/alphagov/pay-connector/pull/5651

src/test/java/uk/gov/pay/api/it/rule/RedisDockerRule.java
```diff
@@ -1,5 +1,7 @@
 package uk.gov.pay.api.it.rule;
 
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
```
DomBelcher, line 7: is this mixing junit 4 and 5? not sure if that matters or not even if it is

DomBelcher: ah ok I see, I didn't know it was possible to use them interchangeably

DomBelcher (approved):
LGTM

src/test/java/uk/gov/pay/api/it/rule/RedisDockerRule.java
oswaldquek: That's the beauty of using interfaces, rather than extending base/abstract classes 🙂

PR #2125, opened 2024-12-05
PP-13367 Rename redis reconnect delay config vars

kbottla:
## WHAT
- Updated `reconnectDelay` config to match AWS docs
- Renamed Redis `reconnectDelay` variables as Full Jitter is not truly exponential. FullJitter delay uses the power of 2 during the calculation and the `base` configured is used to limit the upper bound for a specific attempt. See FullJitter reconnect delay calculation in https://github.com/redis/lettuce/blob/main/src/main/java/io/lettuce/core/resource/FullJitterDelay.java#L59-L63 (FullJitter is explained in AWS docs https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/)

    Reconnect delays calculated locally (with DEBUG enabled and Redis stopped) for the FullJitter approach are as follows:

    ```
      Reconnect attempt 1, delay 100ms
      Reconnect attempt 2, delay 182ms
      Reconnect attempt 3, delay 386ms
      Reconnect attempt 4, delay 516ms
      Reconnect attempt 5, delay 1056ms
      Reconnect attempt 6, delay 3102ms
      Reconnect attempt 7, delay 4092ms
      Reconnect attempt 8, delay 5329ms
      Reconnect attempt 9, delay 9584ms
      Reconnect attempt 10, delay 5446ms
      Reconnect attempt 11, delay 7243ms
      Reconnect attempt 12, delay 9948ms
      Reconnect attempt 13, delay 8209ms
      Reconnect attempt 14, delay 9753ms
      Reconnect attempt 15, delay 6976ms
    ```

    Adds randomness so that no two client nodes (public) try to reconnect simultaneously, causing server (redis) load.

### DIFF
--- a/src/main/java/uk/gov/pay/api/app/config/PublicApiModule.java
+++ b/src/main/java/uk/gov/pay/api/app/config/PublicApiModule.java
@@ -87,9 +87,9 @@ public class PublicApiModule extends AbstractModule {
         ClientResources clientResources = DefaultClientResources.builder()
                 .reconnectDelay(
                         Delay.fullJitter(
-                                configuration.getRedisConfiguration().getExponentialReconnectDelayLowerBound(),
-                                configuration.getRedisConfiguration().getExponentialReconnectDelayUpperBound(),
-                                configuration.getRedisConfiguration().getReconnectDelayExponentBase(),
+                                configuration.getRedisConfiguration().getReconnectDelayLowerBound(),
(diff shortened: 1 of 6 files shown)


jfharden (approved):
Makes sense, much better to have things named in a way that isn't misleading to future people reading the code :)

PR #2144, opened 2025-01-15
PP-13359 Handle AMOUNT_BELOW_MINIMUM

james-peacock-gds:
- Respond appropriately to an AMOUNT_BELOW_MINIMUM response from connector in relation to a create payment request for a Stripe account (we are limiting Stripe payments to 30p or more to avoid the issues associated with fees exceeding the transaction amount).
- Include link to the 'amount' section of the tech docs in error responses for this scenario and for the zero amount not allowed scenario.
- The change to the connector logic to cover this scenario will be made subsequently.


DomBelcher (approved):
LGTM

PR #2149, opened 2025-01-21
PP-13359 Correct 422 message for below minimum amount

james-peacock-gds:
Correct link to docs in error message for a payment below the minimum amount.

### DIFF
--- a/openapi/publicapi_spec.json
+++ b/openapi/publicapi_spec.json
@@ -1558,6 +1558,32 @@
           }
         }
       },
+      "CardDetailsFromResponse" : {
+        "type" : "object",
+        "properties" : {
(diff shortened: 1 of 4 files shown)


openapi/publicapi_spec.json
```diff
@@ -1558,6 +1558,32 @@
           }
         }
       },
+      "CardDetailsFromResponse" : {
```
oswaldquek, line 1561: Why has this been added in this PR?

james-peacock-gds: I have no idea. It occurred when I updated the openapi spec. I assume there was a previous change that wasn't updated correctly

PR #2159, opened 2025-02-11
PP-13334 add a ledger as provider pact test

SandorArpa:
## WHAT YOU DID

Add a pact test that searches Ledger for a transaction which has an honoured corporate exemption. This test is needed because Ledger builds transactions in different ways for a get one and a search call.


hjvoid (approved):
LGTM 👍

PR #2188, opened 2025-04-08
PP-13930: Get refunds from ledger only

oswaldquek:
In practice PaymentRefundsResource never gets refunds from connector, only ledger, so refactor the code to reflect this.


kerrr (approved):
LGTM.
Do we still need the (singular) `GetPaymentRefundStrategy`?

oswaldquek:
> Do we still need the (singular) `GetPaymentRefundStrategy`?

Yes, as just like a charge, the refund may not have been expunged yet.

PR #2202, opened 2025-05-09
PP-14054 Improve validation for some fields

marcotranchino:
With this change, we are improving the input validation for the reference and
description fields so that it matches the same validation currently in place
in SelfService.

### Logging

- [ ] ~~only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.~~

### Documentation

- [ ] ~~Updated README.md for any of the following ?~~

* ~~Introduced any new environment variables / removed existing environment variable~~
* ~~Added new API / updated existing API definition~~

### DIFF
--- a/src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
+++ b/src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
@@ -68,6 +68,7 @@ import static uk.gov.service.payments.commons.model.Source.CARD_PAYMENT_LINK;
 
 class RequestJsonParser {
 
+    public static final char[] NAXSI_NOT_ALLOWED_CHARACTERS = {'<', '>', '|'};
     private static final Set<Source> ALLOWED_SOURCES = EnumSet.of(CARD_PAYMENT_LINK, CARD_AGENT_INITIATED_MOTO);
     public static final Set<AuthorisationMode> ALLOWED_AUTHORISATION_MODES = EnumSet.of(AuthorisationMode.WEB, AuthorisationMode.MOTO_API, AuthorisationMode.AGREEMENT);
 
@@ -199,10 +200,12 @@ class RequestJsonParser {
     }
 
     private static String validateAndGetDescription(JsonNode request, RequestError.Code validationError, RequestError.Code missingFieldError) {
-        return validateAndGetString(
+        String description = validateAndGetString(
                 request.get(DESCRIPTION_FIELD_NAME),
                 aRequestError(DESCRIPTION_FIELD_NAME, validationError, "Must be a valid string format"),
                 aRequestError(DESCRIPTION_FIELD_NAME, missingFieldError));
+        validateNoIllegalCharacters(description, DESCRIPTION_FIELD_NAME, validationError);
+        return description;
     }
 
     private static String validateAndGetPaymentReference(JsonNode paymentRequest) {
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
```diff
@@ -926,6 +933,48 @@ public void getPaymentShouldNotIncludeExemptionWhenNoExemptionThroughLedger() {
                 .body("exemption", is(nullValue()));
     }
 
+    @Test
+    public void shouldReturnBadRequestWhenPaymentReferenceContainsIllegalCharacters() {
+        Map<String, Object> payload = new HashMap<>();
+        payload.put(REFERENCE_FIELD_NAME, ILLEGAL_REFERENCE);
+        payload.put(DESCRIPTION_FIELD_NAME, DESCRIPTION);
+        payload.put(AMOUNT_FIELD_NAME, 1000);
+        payload.put(RETURN_URL_FIELD_NAME, RETURN_URL);
```
alexbishop1, line 942: This could be done with `Map.of(…)`

src/test/java/uk/gov/pay/api/json/RequestJsonParserTest.java
```diff
@@ -130,6 +130,26 @@ void parsePaymentRequest_whenReferenceFieldIsNotAString() throws Exception {
...
 
+    @Test
+    void parsePaymentRequest_whenReferenceFieldContainsIllegalCharacters() throws Exception {
+        for (char illegalChar : RequestJsonParser.NAXSI_NOT_ALLOWED_CHARACTERS) {
+            String payload = "{\n" +
+                    "  \"amount\": 1000,\n" +
+                    "  \"reference\": \"Reference with " + illegalChar + " character\",\n" +
+                    "  \"description\": \"Some description\",\n" +
+                    "  \"return_url\": \"https://somewhere.gov.uk/rainbow/1\"\n" +
+                    "}";
```
alexbishop1, line 141: If there’s ever a time for text blocks, this is it!

alexbishop1 (changes requested):
Looks good. Made a few suggestions.

Since this is technically a behaviour change, I think it would be good to do a quick check of all payments created since, say, the beginning of the month just to check they do not have any disallowed characters in their references or descriptions.

src/test/java/uk/gov/pay/api/it/PaymentsResourceGetIT.java
marcotranchino: Amended, thanks!

src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
```diff
@@ -214,10 +217,22 @@ private static String validateAndGetAgreementReference(JsonNode agreementRequest
...
+        return reference;
+    }
+
+    private static void validateNoIllegalCharacters(String fieldValue, String fieldName, RequestError.Code validationError) {
+        if (fieldValue != null) {
+            for (char c : fieldValue.toCharArray()) {
+                if (NAXSI_NOT_ALLOWED_CHARACTERS.contains(c)) {
+                    throw new BadRequestException(aRequestError(fieldName, validationError, "Must be a valid string format"));
+                }
+            }
```
alexbishop1, line 234: No need to convert it to a character array: `for (int i = 0; i < fieldValue.length(); i++)` (with `fieldValue.charAt(i)` inside `contains(…)` on the next line) would do it. (It would be nice to do this with a stream but unfortunately the `chars()` method on `String` returns an `IntStream` of code points rather than a stream of characters.)

alexbishop1 (changes requested):
Looks good. One small suggestion.

src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
marcotranchino: Amended, thanks Alex! :)

marcotranchino:
I won't merge after approval, waiting for our check tomorrow after standup.

marcotranchino:
I have completed the checks in the production DB noting that there are no payments with reference or description containing illegal characters in the last 6 months, except for payments from one service which has already removed the illegal characters now that we have contacted them.
For this reason, we can safely introduce the change with this PR.

alexbishop1 (approved):

PR #2262, opened 2025-09-16
PP-14410 Accepts and validates property containing recurring payment …

hjvoid:
## WHAT YOU DID
When a new create card payment request has been initiated we need a new property called `agreement_payment_type` which can have a value of `instalment`, `recurring` or `unscheduled`. 

## How to test

- What feedback do you need? 
Not sure if there are adequate tests (outside of `RequestJsonParserTest.java`)?

### DIFF
--- a/src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
+++ b/src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
@@ -2,6 +2,11 @@ package uk.gov.pay.api.json;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.validation.ConstraintViolation;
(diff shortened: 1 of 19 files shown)


src/main/java/uk/gov/pay/api/json/RequestJsonParser.java
```diff
@@ -3,6 +3,7 @@
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.lang3.StringUtils;
+import uk.gov.pay.api.agreement.model.Agreement;
```
SandorArpa, line 6: Unused import

SandorArpa (changes requested):
Some minor comments. It could also do with a rebase to master.
It'd also be nice if you squashed the branches into one.

PR #2387, opened 2026-05-07
PP-15351 Override dependencies

kbottla:
## WHAT YOU DID
- Overrides Jetty and Jackson dependencies due to vulnerable transitive dependencies, and the Dropwizard release with the latest versions of these libraries may take a while.
- Added BOM instead of overriding specific vulnerable dependencies to ensure all modules in the group are aligned to the overridden version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,6 +23,20 @@
     <dependencyManagement>
         <dependencies>
             <dependency>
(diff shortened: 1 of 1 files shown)


Rafiq-Gasmi (approved):
LGTM

PR #2404, opened 2026-06-15
PP-15165 Remove token used to run provider contract tests

kbottla:
## WHAT YOU DID
- Removes `repos_readonly_token` as it is no longer required to run provider contract tests


Rafiq-Gasmi (approved):
LGTM

PR #2419, opened 2026-07-23
PP-15589 remove apache4 dependencies

SandorArpa:
## WHAT YOU DID

Refactor to remove Apache v4 dependencies from the code.


src/test/resources/logback.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
     <logger name="org.apache.http" level="info" />
+    <logger name="org.apache.hc.client5.http" level="WARN"/>
```
AndrzejProchyra, line 4: Should this go in a separate PR?

SandorArpa: No. This is needed since all the test classes are using client5.

AndrzejProchyra: OK, because, as you have explained it to me, otherwise it does not log during tests and we want it to.

AndrzejProchyra (approved):

PR #2449, opened 2026-09-15
[PP-15703] Add workflow to generate transitive dependency graph for dependabot

foxmaltas-gds:
## WHAT YOU DID
Added a github actions workflow to generate a java transitive dependency graph to allow us to keep the transitive dependencies up to date

## How to test
This has been tested on `pay-connector` and is working. It will run once this PR is merged

### Logging

- [x] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### Documentation

- [x] Updated README.md for any of the following ?


SandorArpa (approved):
LFTM

